PR #24, opened 2021-09-01
I find a better plugin in clipboard

1096015776:
I always use clipboard
but sometime , the clipboard is not friendly for me
i find a clipboard use telescope to brower my clipboard history

and i want to some text just delete not put in registers
so i use backspace like i begin use editor 
maybe this is friendly for newer vim user

i try it become let this become better
because i really like this nvim config

### DIFF
--- a/lua/keymap/init.lua
+++ b/lua/keymap/init.lua
@@ -89,6 +89,8 @@ local plug_map = {
         :with_silent(),
     ["n|<Leader>fg"] = map_cu('Telescope git_files'):with_noremap()
         :with_silent(),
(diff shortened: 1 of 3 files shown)


ayamir:
I use [clipmenud](https://github.com/cdown/clipmenu) and [copyq](https://github.com/hluk/CopyQ) to manager my clipboard.
I don't know how this experience is convenient. How about your own feeling?

1096015776:
I not experience clip menu ,maybe it is better for just at nvim.
because clib menus should be glboal.
I want to experience that you said menu

1096015776:
thanks for that answer
the neo cliblike git file just fzf search maybe not different from clibmenu
so it is good for me sometime

PR #47, opened 2021-11-26
fix symbols-outline wide error

HATTER-LONG:
Minimap causes the symbol list to display abnormally, adjust to the left。

or just fix symbol-outline config not take effect. 

REF:[issue](https://github.com/simrat39/symbols-outline.nvim/issues/62)


ayamir:
fixed at 95d1c8d82986c5a44096afb269901da15e9400fd.

PR #106, opened 2022-04-22
Use cmp-cmdline and illuminate to improve the use experience

feng2r200:
These two plugins are written in Lua and have good compatibility with CMP and LSP. They provide more accurate completion and highlighting, and do not affect the use experience when large files are used

### DIFF
--- a/lua/modules/completion/config.lua
+++ b/lua/modules/completion/config.lua
@@ -312,6 +312,22 @@ function config.cmp()
 			-- {name = 'cmp_tabnine'}
 		},
 	})
+
+    cmp.setup.cmdline("/", {
+        mapping = cmp.mapping.preset.cmdline(),
+        sources = { { name = "buffer" } }
+    })
+    cmp.setup.cmdline("?", {
+        mapping = cmp.mapping.preset.cmdline(),
+        sources = { { name = "buffer" } }
+    })
+    cmp.setup.cmdline(":", {
+        mapping = cmp.mapping.preset.cmdline(),
+        sources = cmp.config.sources(
+            { { name = "path" } },
+            { { name = "cmdline" } }
+        )
+    })
 end
 
 function config.luasnip()
--- a/lua/modules/completion/lsp.lua
+++ b/lua/modules/completion/lsp.lua
@@ -5,6 +5,7 @@ vim.cmd([[packadd lsp_signature.nvim]])
 vim.cmd([[packadd lspsaga.nvim]])
 vim.cmd([[packadd cmp-nvim-lsp]])
 vim.cmd([[packadd aerial.nvim]])
+vim.cmd([[packadd vim-illuminate]])
 
 local nvim_lsp = require("lspconfig")
 local saga = require("lspsaga")
@@ -63,6 +64,7 @@ local function custom_attach(client)
 		handler_opts = { "double" },
 	})
 	require("aerial").on_attach(client)
+    require("illuminate").on_attach(client)
 
 	if client.resolved_capabilities.document_formatting then
 		vim.cmd([[augroup Format]])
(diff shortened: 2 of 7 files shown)


flaviusbuffon:
Wondering the reason for using plugin cmp-cmdline instead of wilder?

feng2r200:
> Wondering the reason for using plugin cmp-cmdline instead of wilder?

Personally, I think that in practical use, the implementation of incomplete command line functions is similar, but CMP CmdLine is based on CMP, which has better compatibility, less code and better stability. In addition, it can obtain the support of other CMP plug-ins by adding source.And I personally prefer to use fewer plug-ins, because more plug-ins and more file scanning. VIM does not respond well when opening large files of large projects, so I will try to simplify the plug-ins as much as possible (my company project has a lot of large Java code, and the redrawing page will get stuck). CMP CmdLine is also a pull request I decided to launch after a long time. May I know why you like Wilder? I'll take the time to look at their code and learn more about it

ayamir:
A new bug: https://www.reddit.com/r/neovim/comments/ucze6k/q_any_idea_why_i_cannot_do_tab_completion_on/
So I will roll back to wilder util it is fixed.

feng2r200:
> A new bug: https://www.reddit.com/r/neovim/comments/ucze6k/q_any_idea_why_i_cannot_do_tab_completion_on/
> So I will rollback to wilder util it fixed.

I saw this. I also encountered this problem some time ago, but now there is no such problem.At least I can't reproduce this problem for now.

flaviusbuffon:
I have also encountered this problem, and I believe the reasons are as follows. cmp-cmdline depends on nvim-cmp to work, and nvim-cmp sets event = "InsertEnter", so you have to return to command mode after editing to achieve command completion.
The reddit post just noticed it can not used in command mode sometimes, but may be dismissed the essence of this plugin.

Also, that is a reason for why I want to use wilder instead of cmp-cmdline.

feng2r200:
> I have also encountered this problem, and I believe the reasons are as follows. cmp-cmdline depends on nvim-cmp to work, and nvim-cmp sets event = "InsertEnter", so you have to return to command mode after editing to achieve command completion. The reddit post just noticed it can not used in command mode sometimes, but may be dismissed the essence of this plugin.
> 
> Also, that is a reason for why I want to use wilder instead of cmp-cmdline.

I see. I'll switch my configuration to wilder and follow up their repostories

flaviusbuffon:
> > I have also encountered this problem, and I believe the reasons are as follows. cmp-cmdline depends on nvim-cmp to work, and nvim-cmp sets event = "InsertEnter", so you have to return to command mode after editing to achieve command completion. The reddit post just noticed it can not used in command mode sometimes, but may be dismissed the essence of this plugin.
> > Also, that is a reason for why I want to use wilder instead of cmp-cmdline.
> 
> I see. I'll switch my configuration to wilder and follow up their repostories

I think it can not be viewed as a problem carried by `cmp-cmdline`, which is inevitable! May be you can try to solve this problem by change `nvim-cmp` trigger `event `setting. But please be carefully since the dependencies and event handling of `packer` are not very well done

feng2r200:
> > > I have also encountered this problem, and I believe the reasons are as follows. cmp-cmdline depends on nvim-cmp to work, and nvim-cmp sets event = "InsertEnter", so you have to return to command mode after editing to achieve command completion. The reddit post just noticed it can not used in command mode sometimes, but may be dismissed the essence of this plugin.
> > > Also, that is a reason for why I want to use wilder instead of cmp-cmdline.
> > 
> > 
> > I see. I'll switch my configuration to wilder and follow up their repostories
> 
> I think it can not be viewed as a problem carried by `cmp-cmdline`, which is inevitable! May be you can try to solve this problem by change `nvim-cmp` trigger `event `setting. But please be carefully since the dependencies and event handling of `packer` are not very well done

Thanks. I have found the reason why I have no problems here. A very important content has been omitted from the code I submitted, which is mentioned above

image

The event was not changed when submitting

PR #125, opened 2022-05-16
feat: Add brglng/vim-im-select to change input method automatically.

cherrot:
related issue: #112 

have tested on my Linux (with `fcitx5` installed) and MacOS (need install `im-select` first)


cherrot:
I'll check if this is a bug, thx

PR #126, opened 2022-05-16
fix: Add file_browser extension and fix colorscheme conflict

cherrot:
## colorscheme conflict
I'm not sure if this is a bug. In `config.edge()` there is one line to set a light background:
```
vim.cmd([[set background=light]])
```

This line would break settings in `nord` and give a weird color scheme when you switch to `nord`.

## redundant config?
I guess in `config.indent_blankline()` the following lines are redundant? 
```
vim.opt.termguicolors = true
vim.opt.list = true
```

## Add missing dependency
`nvim-telescope/telescope-file-browser.nvim` is needed in key-mapping. Again, I'm not sure whether I should add it back 😉


ayamir:
I forgot to delete this keymap, no need to readd it because nvim-tree can replace it well.

ayamir:
I have fixed it manually, thanks for your pr.

PR #166, opened 2022-07-27
Consider replace gitui with lazygit

jczhang02:
`gitui` do not support signed commit, while `lazygit` supports


ayamir (approved):
OK, lazygit is a better option.

PR #167, opened 2022-07-27
Clear neovim python interpreter setting

jczhang02:
Generally speaking, some packages needed by `neovim` are installed in the system default python environment, but the `check_conda()` function sets the `neovim` python path to the python in the `conda` environment, which can easily cause errors for some plugins.


ayamir (approved):
ok, you are right.

PR #169, opened 2022-07-27
Fix warning of dap plugins

jczhang02:
- Delete `dap-buddy` plugin since mason.nvim
- Adjust dap plugins order


ayamir (approved):
ok, fix problem and remove redundant plugin.

PR #174, opened 2022-08-05
Use toggleterm for horizontal,vertical and float terminals, plus lazygit ;)

cherrot:
I found `toggleterm.nvim` is functional for all the use cases in this config. So I remove `split-term` and `FTerm.nvim`, leaving `toggleterm.nvim` the only terminal wrapper. Details below:

- Although not being used, I've enhanced `pbind.nvim_load_mapping(mapping)` to allow setting keymap on a specific buffer (`0` represents the current)
- Add global function `_G.toggle_lazygit` to toggle the `lazygit` floating window.
- ~~We are not able to switch to normal mode (with `<C-\><C-n>`) in terminal if we use the default keybinding `<C-\>` to toggle the horizontal terminal. So I change it to `<C-=>`, `=` is next to `\`, easy to get used to~~
- Make the plugin being loaded on "UIEnter" event, in case we want to use the terminal before editing any file.

### DIFF
--- a/lua/keymap/bind.lua
+++ b/lua/keymap/bind.lua
@@ -9,6 +9,7 @@ function rhs_options:new()
 			expr = false,
 			nowait = false,
 		},
+		buffer = false,
 	}
 	setmetatable(instance, self)
 	self.__index = self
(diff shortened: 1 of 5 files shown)


ayamir:
It's great that you reduce the plugins. But you'd better make the keybinding compatiable with original.

cherrot:
I've reverted to the default keybindings except `<C-\>`. Do you have any suggestions on it?

ayamir:
`<C-=>` is actually some terminal's keybinding for increase font size, it's not a good idea.

ayamir:
image
It works.

cherrot:
Done :)

ayamir (approved):

PR #187, opened 2022-08-14
chore(nvim-tree): Update options from upstream.

Jint-lzxy:
This commit update [nvim-tree](https://github.com/kyazdani42/nvim-tree.lua)'s options from upstream. Some configs are modified from nvim-tree's recent commits to ensure that this configuration still works correctly. _(For example, `folder_arrow = false` is set to ensure proper alignment)_. However, there are no changes to the existing options _(i.e., the appearance does not change)_.

There're also several cleanups, that's, for example, using `default` instead of `["default"]` for consistency _(L668-690)_

> See [:help nvim-tree-setup](https://github.com/kyazdani42/nvim-tree.lua/blob/master/doc/nvim-tree-lua.txt) for changes and options in detail.

### DIFF
--- a/lua/modules/ui/config.lua
+++ b/lua/modules/ui/config.lua
@@ -599,6 +599,7 @@ end
 
 function config.nvim_tree()
 	require("nvim-tree").setup({
+		create_in_closed_folder = false,
 		respect_buf_cwd = true,
 		auto_reload_on_write = true,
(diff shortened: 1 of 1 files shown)


Jint-lzxy:
Current Layout:

Layout

ayamir (approved):

PR #189, opened 2022-08-15
feat: Use nvim-navic for showing code context.

### LINKED ISSUE #188: Consider using nvim-navic and aerial.nvim?
- For [nvim-navic](https://github.com/SmiteshP/nvim-navic), it's because [nvim-gps](https://github.com/SmiteshP/nvim-gps) has been deprecated, and from the project description:
> nvim-navic is an improved version of nvim-gps (a version 2.0!). It accomplishes the same task as this plugin, but uses LSP instead of Treesitter. It also adds ability to have coloured icons!

- However, there's a frequently reported issue: `"x does not support documentSymbols" error popping up every time an editor is opened`. The standard recommendation is to not attach navic to problematic servers. While this is reasonable in some cases, there are a number of servers that navic would be useful with, like `htmlls` or `cssls`, that still use the `SymbolInformation` interface. **So shall we switch to nvim-navic once `SymbolInformation` is supported? There is a PR attempting to address this problem now. _(Another choice - which might be better - is to load both plugins and set gps as a fallback when navic is unavailable under current buffer)._**
******
- Given that the `SymbolInformation` API is still in use by a number of existing servers, the same applies to [lspsaga.nvim](https://github.com/glepnir/lspsaga.nvim) _(Specifically, the **outline** feature [[this line]](https://github.com/glepnir/lspsaga.nvim/blob/bd95871d67b8942a7869f8284c151b85f653027b/lua/lspsaga/outline.lua#L14))_. Whereas [aerial.nvim](https://github.com/stevearc/aerial.nvim) could automatically fetch symbols from Treesitter as a backend, with proper configuration it could offer the same experience as lspsaga **and** provide support for a wide range of languages at the same time _(Now some languages (and servers spawned by efm) cannot view it's symbol outline, lspsaga would throw an error)_. **Should we add aerial.nvim back again?**

Jint-lzxy:
As mentioned in #188. This PR would also close #188 once merged.

- Note that we replace `nvim-gps` with `nvim-navic` in _most_ cases, but gps is still preserved for Treesitter fallback _(i.e., for filetypes / servers that are unsupported by nvim-navic)_
- Catppuccin's integration with navic is also enabled for nice-looking highlights.
- The icons for navic came from nvim-cmp.

**Minor Change**
- Disabled highlight settings for aerial of catppuccin

### DIFF
--- a/lua/modules/completion/lsp.lua
+++ b/lua/modules/completion/lsp.lua
@@ -4,9 +4,9 @@ vim.cmd([[packadd lsp_signature.nvim]])
 vim.cmd([[packadd lspsaga.nvim]])
 vim.cmd([[packadd cmp-nvim-lsp]])
 vim.cmd([[packadd vim-illuminate]])
+vim.cmd([[packadd nvim-navic]])
 
 local nvim_lsp = require("lspconfig")
-local saga = require("lspsaga")
 local mason = require("mason")
 local mason_lsp = require("mason-lspconfig")
 
@@ -25,7 +25,7 @@ mason_lsp.setup({
 local capabilities = vim.lsp.protocol.make_client_capabilities()
 capabilities = require("cmp_nvim_lsp").update_capabilities(capabilities)
 
-local function custom_attach(client)
+local function custom_attach(client, bufnr)
 	require("lsp_signature").on_attach({
 		bind = true,
 		use_lspsaga = false,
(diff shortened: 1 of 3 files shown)


Jint-lzxy:
**Showcase:**
- navic _(ft: lua)_:
navic

- gps _(ft: shell script, fallback)_:
gps

ayamir (approved):

PR #204, opened 2022-08-24
chore(catppuccin): Update options from upstream.

Jint-lzxy:
## Fast-forward
- `nvim-tree` doesn't require nested options any more.
- Enabled nested options for `bufferline`.
- Update hl group `bufferline` so that the highlights stay consistent with `native-lsp`'s configs.


Jint-lzxy:
Current style _(bufferline)_:
bufferline-demo

ayamir (approved):

PR #212, opened 2022-08-27
fix(catppuccin): Integration with navic.

Jint-lzxy:
If `custom_bg` was not set then we could get a strange-looking white background or no highlight.


nullchilly:
@Jint-lzxy I also moved bufferline to respect the author's guidelines, please update it too!

More info here: https://github.com/catppuccin/nvim/issues/260#issuecomment-1229178598

ayamir (approved):

PR #238, opened 2022-09-28
fix: invalid path in dap python configuration.

Lraxer:
DAPInstall was replaced by mason and removed in #169 , but the python configuration of nvim-dap stays unchanged. 

The official tutorial [Debug Adapter installation](https://github.com/mfussenegger/nvim-dap/wiki/Debug-Adapter-installation#python) recommends using virtual environment, but I think the easiest way is either applying [nvim-dap-python](https://github.com/mfussenegger/nvim-dap-python) or using `/usr/bin/python` by `pip install debugpy` directly. In this PR I use the latter and it works fine in archlinux. What do you think of it?

BTW [wiki](https://github.com/ayamir/nvimdots/wiki/Plugins#editor) also needs to update, but it seems that only repo owner can edit.

### DIFF
--- a/lua/modules/editor/config.lua
+++ b/lua/modules/editor/config.lua
@@ -349,7 +349,7 @@ function config.dap()
 
 	dap.adapters.python = {
 		type = "executable",
-		command = os.getenv("HOME") .. "/.local/share/nvim/dapinstall/python/bin/python",
+		command = "/usr/bin/python",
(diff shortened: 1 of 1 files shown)


ayamir (approved):

ayamir:
ok, I removed the restriction so that you can try editing it.

PR #240, opened 2022-09-30
make copilot works under cmp framework, use lspkind to provide icons.

ayamir:


ayamir:
@fleongg Can you review it?

Issue #168, opened 2022-07-27
能否加入不配置ssh key方式安装？


MoxiuHub:
如图，因特殊安全问题不能配置ssh key 希望改为/加入此方式～

<sub>Sent from <a href="https://apps.apple.com/cn/app/id1314212521">PPHub</a></sub>

ayamir:
你可以自己手动修改：
把下面这几行中的`git@github.com:`改成`https://github.com/`即可
https://github.com/ayamir/nvimdots/blob/323f1df0582d61eea1c3fe61f848f9e6035a433d/lua/core/pack.lua#L42
https://github.com/ayamir/nvimdots/blob/323f1df0582d61eea1c3fe61f848f9e6035a433d/lua/core/pack.lua#L53
https://github.com/ayamir/nvimdots/blob/323f1df0582d61eea1c3fe61f848f9e6035a433d/lua/core/pack.lua#L76

jczhang02:
This issue can be closed.

PR #263, opened 2022-10-07
chore(catppuccin): Update compile fn.

Jint-lzxy:
> Ref: https://github.com/catppuccin/nvim/pull/289

### DIFF
--- a/lua/core/init.lua
+++ b/lua/core/init.lua
@@ -105,8 +105,8 @@ local load_core = function()
 	require("core.event")
 	pack.load_compile()
 
-	-- vim.cmd([[set background=light]])
-	vim.cmd([[colorscheme catppuccin]])
+	-- vim.api.nvim_command([[set background=light]])
+	vim.api.nvim_command([[colorscheme catppuccin]])
 end
 
 load_core()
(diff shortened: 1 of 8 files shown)


nullchilly:
autocmd and packer hook is no longer required

Jint-lzxy:
@nullchilly Thanks for the heads up :)

nullchilly:
I wanted to try this config out but the packer bootstrap seems broken?

image

Jint-lzxy:
@nullchilly This is due to the original design pattern _(because the direct use of `git clone` in Mainland China will have "connection refused" as response in most cases, so @ayamir switched to ssh)_. More discussion at https://github.com/ayamir/nvimdots/issues/49 and [wiki](https://github.com/ayamir/nvimdots/wiki/Prerequisites). The error message `Host key verification failed` indicates that `ssh` is not configured correctly. Alternatively, I'm migrating the installation process to the direct use of an installation script to detect problems. I'm so sorry that this has troubled you.

I will add `git` installation method to the install script recently (similar issue: https://github.com/ayamir/nvimdots/issues/168). Current command:
```sh
/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/ayamir/nvimdots/0.8/install/install.sh)"
```

ayamir (approved):

PR #279, opened 2022-10-14
feat: update nvim-colorizer.lua to a new fork

CharlesChiuGit:
For the old one is not maintained. And the NvChad one had included some important merge.


Jint-lzxy (approved):
LGTM 👍

PR #283, opened 2022-10-14
feat: user can set is_sync_with_ssh from installation and settings.lua

ayamir:


ayamir:
@Jint-lzxy I modified the `install.sh` to make user can choose which method to sync plugins, and set a bool variable `is_sync_with_ssh` to be used in any lua file which need it.

Jint-lzxy:
@ayamir I modified some part of the code [\[view diff here\]](https://github.com/ayamir/nvimdots/compare/5a496e6e4a79973082e580f5d3f34460245720bb..7f0fd678def3982cf07f7871ae1a9c2600f231b9), including:
- On the OSX version of `sed`, the `-i` option requires **mandatory** provision of the **extension parameter**. _(In your code, the original replacement command is actually parsed as an extension parameter, while the file path is interpreted as the next command line parameter on OSX.)_
```diff
-  sed -i 's/\[\"use_ssh\"\] \= true/\[\"use_ssh\"\] \= false/g' ./lua/core/settings.lua
+  sed -i '' -e 's/\[\"use_ssh\"\] \= true/\[\"use_ssh\"\] \= false/g' "./lua/core/settings.lua"
```
- `is_sync_with_ssh -> use_ssh`, `default_url_format -> clone_prefix` more intuitive name.
- **Squashed** two commits into a single one.
- **Replace** the original `if` statement with more concise statements, like:

From:
```lua
local clone_prefix = "git@github.com:%s"
if not use_ssh then
	clone_prefix = "https://github.com/%s"
end
```
To:
```lua
local clone_prefix = use_ssh and "git@github.com:%s" or "https://github.com/%s"
```

Jint-lzxy:
Another change: We don't need to validate `SSH` for Treesitter, `prefer_git` is just the difference between `curl` and `git`, not `HTTPS` and `SSH`.

ayamir:
ok, approve and merge it.

Jint-lzxy (approved):

PR #288, opened 2022-10-16
chore(settings): use home variable

CharlesChiuGit:


ayamir (approved):

CharlesChiuGit (commented):
@ayamir Pure curious, is using `os.getenv("HOME")` faster? Or if there's any reason you do it this way?(I'm totally fine with it, just wonder why.)

ayamir:
https://github.com/nanotee/nvim-lua-guide/issues/50#issuecomment-783579296
`vim.env` is a wrapper for `getenv()` and `setenv()`.

CharlesChiuGit:
Got it! Thanks for the info!

PR #308, opened 2022-10-28
fix(core.pack): add missing space

CharlesChiuGit:


CharlesChiuGit:
@Jint-lzxy it's a small bug, but it cause installation problem via https.

Jint-lzxy (approved):
Thanks :)

PR #311, opened 2022-11-04
feat(script): Support cross-platform installation on *nix.

Jint-lzxy:
**This commit updated the installation script to ensure cross platform-support _(on \*nix)_.**

The details are as follows:
- All pathnames and variables are now enclosed in quotation marks to prevent installation failure due to space(s) in variables.
- **!!** **Replace** the `sed` command with `perl`. AFAIK `perl` is now provided on `MacOS`, `Unix`, and `Linux` _(Linux reference: [LSB Languages Version 5.0](https://refspecs.linuxfoundation.org/LSB_5.0.0/LSB-Languages/LSB-Languages/perllocation.html) and [Download page for Perl](https://www.perl.org/get.html))_. This can address compatibility issues of the `sed` command with `-i` option.

@ayamir @CharlesChiuGit Can you help me test this on Linux? I don't have machines running Linux at hand, but AFAIK `perl` _should_ work correctly on every distribution 👍

```sh
/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Jint-lzxy/nvimdots/script-compatibility/install/install.sh)"
```

### DIFF
--- a/install/install.sh
+++ b/install/install.sh
@@ -123,13 +123,22 @@ check_ssh() {
 
 is_latest() {
 	local nvim_version
-	nvim_version=$(nvim --version | head -n1 | sed -e 's|^[^0-9]*||' -e 's| .*||')
+	nvim_version="$(nvim --version | head -n1 | sed -e 's|^[^0-9]*||' -e 's| .*||')"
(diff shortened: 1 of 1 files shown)


CharlesChiuGit:
Cool! Let me test it.

ayamir (approved):

CharlesChiuGit:
Sorry for the late reply, the script works fine on Ubuntu22.04.

PR #346, opened 2022-12-04
chore(stylua-action): bump to v2

CharlesChiuGit:


Jint-lzxy (approved):

CharlesChiuGit:
@Jint-lzxy The author said:
> NOTE: using `latest` may cause the action to fail if StyLua updates and the formatting changes!
I thought it's better not to include the version tag?

Jint-lzxy:
> I thought it's better not to include the version tag?

@CharlesChiuGit If the version tag is **not** specified, `latest` will still be used as a fallback. For example: [style check for #346](https://github.com/ayamir/nvimdots/actions/runs/3611471293/jobs/6085976976#step:3:6).

If the upstream formatting rules change, we will make adjustments accordingly 👍

PR #352, opened 2022-12-10
chore(catppuccin): Update highlights

Jint-lzxy:
Due to changes in the upstream _(nvim-treesitter)_ highlight groups, I have also modified several highlights for catppuccin to prevent duplication in adjacent highlight groups _(in some cases, they cannot be distinguished at a glance)_, for example:
```cpp
                 static constexpr                          foo                     <int> bar;
                       ↑                                    ↑                        ↑
storage_class_specifier/type_qualifier(yellow)    type_identifier(yellow)   primitive_type(yellow, italic)
```
has been adjusted to:
```cpp
           static constexpr                          foo                     <int> bar;
                 ↑                                    ↑                        ↑
storage_class/type_qualifier(red, italic)    type_identifier(yellow)   primitive_type(yellow, italic)
```
******
### This modification is _non-trivial_. It will affect the syntax highlighting of the following languages:
- blueprint
- C
- C#
- C++
- CSS
- cuda
- D
- dart
- fortran
- gdscript
- gleam
- glsl
- hack
- hlsl
- java
- julia
- kotlin
- lalrpop
- llvm
- mermaid
- ocaml
- pascal
- php
- pioasm
- ql
- qmljs
- rasi
- ruby
- rust
- scala
- solidity
- swift
- typescript
- vala
- verilog
- wgsl
- zig

### DIFF
--- a/lua/modules/ui/config.lua
+++ b/lua/modules/ui/config.lua
@@ -246,6 +246,7 @@ function config.catppuccin()
 					Keyword = { fg = cp.pink },
 					Type = { fg = cp.blue },
 					Typedef = { fg = cp.yellow },
+					StorageClass = { fg = cp.red, style = { "italic" } },
 
 					-- For native lsp configs.
 					DiagnosticVirtualTextError = { bg = cp.none },
@@ -285,6 +286,7 @@ function config.catppuccin()
 					["@constant.builtin"] = { fg = cp.lavender },
 					-- ["@function.builtin"] = { fg = cp.peach, style = { "italic" } },
 					-- ["@type.builtin"] = { fg = cp.yellow, style = { "italic" } },
+					["@type.qualifier"] = { link = "@keyword" },
 					["@variable.builtin"] = { fg = cp.red, style = { "italic" } },
 
 					-- ["@function"] = { fg = cp.blue },


ayamir (approved):

nullchilly:
@Jint-lzxy Hey would you be up to be a collaborator of https://github.com/catppuccin/nvim/

A lot of good color tweaks goes to waste because it was not PRed to upstream (For example C++ highlights are much better here)

Thanks :), Null Chilly

Jint-lzxy:
Sorry for the late reply, it's getting closer to the end of the year and I'm all wound up recently.

> @Jint-lzxy Hey would you be up to be a collaborator of https://github.com/catppuccin/nvim/
> 
> A lot of good color tweaks goes to waste because it was not PRed to upstream (For example C++ highlights are much better here)
> 
> Thanks :), Null Chilly

@nullchilly Great :) Recent commits to nvim-treesitter have brought way more refined hl groups, and IMO simple linking may result in similar highlight groups with different functionalities. I will look into this next year 👍

nullchilly:
Invitation sent, thanks you :)

I will make the best out of 2022 too!

PR #371, opened 2022-12-18
feat(core): hot reload nvim core options

CharlesChiuGit:
This autocmd can hot-reload nvim configs when core options being modified. 
Plugins config may or may not work, depends on how plugin read their config.

NOTE: the `~/.config/nvim` must not be a sym-link.

### DIFF
--- a/lua/core/event.lua
+++ b/lua/core/event.lua
@@ -15,15 +15,14 @@ end
 
 function autocmd.load_autocmds()
 	local definitions = {
-		packer = {
+		packer = {},
+		bufs = {
 			-- Hot reload nvim core config
 			{
 				"BufWritePost,FileWritePost",
 				"~/.config/nvim/*.lua",
 				[[source % | source $MYVIMRC | redraw! | PackerCompile]],
 			},
-		},
-		bufs = {
 			{ "BufWritePre", "/tmp/*", "setlocal noundofile" },
 			{ "BufWritePre", "COMMIT_EDITMSG", "setlocal noundofile" },
 			{ "BufWritePre", "MERGE_MSG", "setlocal noundofile" },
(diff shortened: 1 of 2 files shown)


lua/core/event.lua
```diff
@@ -15,19 +15,15 @@ end
...
-		packer = {},
-		bufs = {
-			-- Reload vim config automatically
-			{
-				"BufWritePost",
-				[[$VIM_PATH/{*.vim,*.yaml,vimrc} nested source $MYVIMRC | redraw]],
-			},
-			-- Reload Vim script automatically if setlocal autoread
+		packer = {
+			-- Hot reload nvim core config
```
Jint-lzxy, line 19: Why put this autocmd under the `packer` table? This event does not belong to the `Packer` augroup

Jint-lzxy (changes requested):
Is it necessary to avoid reminding the user `Packer` has been recompiled every time the files are written? ... and I'm suspicious that this event is placed in the right place.

lua/core/event.lua
CharlesChiuGit: Make sense, will put it in `bufs` augroup

CharlesChiuGit:
I'm also thinking about the same thing, the notify is a bit annoying. But I thought users will only get notify if their editing the nvim config, maybe it won't be too disturbing?

Jint-lzxy:
@CharlesChiuGit I figured out a workaround, will push the change now

Jint-lzxy:
Basically this commit added an optional parameter that can disable notify when passed in.

CharlesChiuGit:
Nice! Thanks a lot!

Jint-lzxy (approved):

Jint-lzxy:
LGTM Now 👍

ayamir (approved):

Issue #375, opened 2022-12-19
Questions about `lsp.lua` implementation details


Jint-lzxy:
### Neovim version

NVIM v0.8.1

### Operating system/version

macOS 12.6.2 (21G320) [x86_64]

### Terminal name/version

iTerm2 Build 3.4.18

### $TERM environment variable

xterm-256color

### Branch info

main (Default/Latest)

### Fetch Preferences

SSH (use_ssh = true)

### How to reproduce the issue

I found three potential issues while reading the code related to LSPs:
1. When setting up `clangd`, as the essence of table assignment in Lua is to pass references _(equivalent to `LUA_TTABLE&` in C++)_, the `copy_capabilities` table does not do anything actually. Any language servers that are set up later will have `offsetEncoding = { "utf-16" }`. I don't think this is the expected result:
https://github.com/ayamir/nvimdots/blob/2f04ac102f53c4147b9d0190318aa335af8ce64c/lua/modules/completion/lsp.lua#L108-L110
2. @CharlesChiuGit The file `lsp.lua` will also be reloaded during hot-reload, which means that all language servers will restart every time a file is saved. This issue might need to be addressed very soon.
3. In lua, if a function needs no more parameters and that these extra parameters are at the end of the function, it is redundant to replace them with placeholder _(`_`)_. So we could just delete these placeholders, for example:
https://github.com/ayamir/nvimdots/blob/0d818ba9175bd720827a6324036039579989360f/lua/modules/editor/config.lua#L331

### Expected behavior

1. Server capabilities are copied correctly
2. Servers will not restart during hot-reload
3. Remove unnecessary code

CharlesChiuGit:
Can confirm.

CharlesChiuGit:
@Jint-lzxy Would it be a more ideal way if we hot-reload based on each folder?
Since if we change some configs about lsps, we might also want it to be reloaded.
But code would looks much bloated.

Jint-lzxy:
@CharlesChiuGit IMO we can provide user command _(or keybinding)_ for users to do hot-reload. Hot-reload on every save may not be the most efficient approach

ayamir:
> @CharlesChiuGit IMO we can provide user command _(or keybinding)_ for users to do hot-reload. Hot-reload on every save may not be the most efficient approach

I don't think `PackerCompile` after everytime `*.lua` saved is a good idea. I usually save config frequently so this function always cause error like this:
[...]

PR #376, opened 2022-12-19
fix(lsp): use clang default capabilities with utf-16 as the first offsetEncoding option.

ayamir:
I found the default capabilities here: 
https://github.com/neovim/nvim-lspconfig/blob/973aa14d0992df82ff82f714d978a3eb8d676600/lua/lspconfig/server_configurations/clangd.lua#L33
#375


Jint-lzxy (changes requested):
IMO we should **append** to default capabilities instead of _completely_ replacing them, because the default capabilities also include necessary settings for nvim-cmp

Jint-lzxy (approved):
@ayamir Fixed 👍

PR #377, opened 2022-12-19
feat: User-command "HotReload"

Jint-lzxy:
This commit modularizes the hot-reload feature. The user can execute hot-reload through `<leader>R` or `:HotReload`. 

@CharlesChiuGit Can you help me test whether the expected result is achieved?

### DIFF
--- a/lua/core/init.lua
+++ b/lua/core/init.lua
@@ -69,6 +69,13 @@ local disable_distribution_plugins = function()
 	-- vim.g.loaded_remote_plugins = 1
 end
 
+local set_hot_reload = function()
+	vim.api.nvim_create_user_command("HotReload", function()
+		require("core.pack").back_compile(true)
+		vim.api.nvim_command([[source $MYVIMRC | redraw! | echom "Hot-reload success!"]])
+	end, { force = true })
+end
+
 local leader_map = function()
 	vim.g.mapleader = " "
 	vim.api.nvim_set_keymap("n", " ", "", { noremap = true })
@@ -117,6 +124,7 @@ local load_core = function()
 	local pack = require("core.pack")
 	createdir()
 	disable_distribution_plugins()
+	set_hot_reload()
 	leader_map()
 
 	pack.ensure_plugins()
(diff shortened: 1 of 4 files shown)


CharlesChiuGit:
wow, so fast!

ayamir:
Make it a configurable setting at `settings.lua` and default equals `false` is more reasonable.

CharlesChiuGit:
Maybe we can also revert the workaround for disable notify on `PackerCompile`?

lua/core/init.lua
```diff
@@ -69,6 +69,13 @@ local disable_distribution_plugins = function()
 	-- vim.g.loaded_remote_plugins = 1
 end
 
+local set_hot_reload = function()
+	vim.api.nvim_create_user_command("HotReload", function()
+		require("core.pack").back_compile(true)
+		vim.api.nvim_command([[source $MYVIMRC | redraw! | echom "Hot-reload success!"]])
```
CharlesChiuGit, line 75: we also need to source the current changed file, so the changes will be valid.

Jint-lzxy: Then we would need to judge whether the currently saved file is legal? _(whether it is in the correct path and whether the extension is `*.lua`)_

Jint-lzxy: I don't think it has been resolved yet. `neovim` will also source extraneous files, we should check before execute

CharlesChiuGit: yeah, i think it will be necessary.

CharlesChiuGit: > @CharlesChiuGit I don't think it has been resolved yet. `neovim` will also source extraneous files, we should check before execute

Sorry, will do.

Jint-lzxy:
I agree with @ayamir. Frequent reloads can cause some potential problems, and many plugins actually don't support hot reloading their settings when they are `required` again. Gonna close this now, what's your opinion? @CharlesChiuGit @ayamir

ayamir:
> Gonna close this now

We need to revert corresponding commits.

ayamir:
> Ok, sorry for causing troubles.

No need to say sorry, just be careful next time. :)

Jint-lzxy:
> Ok, sorry for causing troubles.

Yeah. A very good feature, but not _that_ practical to implement it 👍

PR #383, opened 2022-12-24
perf(ui): ui twick & update theme-related config

CharlesChiuGit:
Add some nice visual effects based on current plugins.
1. `wilder`
wilder

2. `indent-blacnkline`
indent-line

3. `virtual-text`
v-text


ayamir:
I have just be diagnosed with COVID-19 and need to rest for a few days.

lua/modules/tools/config.lua
```diff
@@ -225,8 +225,14 @@ function config.wilder()
 		),
 	})
 
+	local match_hl = require("utils").hlToRgb("String", false)
```
Jint-lzxy, line 228: In this case, we no longer need the `utils` module. This line can be replaced as follows:
```suggestion
	local string_fg = vim.api.nvim_get_hl_by_name("String", true).foreground
	local match_hl = string_fg ~= nil and string.format("#%06x", string_fg) or "#ABE9B3"
```
Note that `String` is not one of the builtin hl groups. We need to set a fallback value for it.

Jint-lzxy (changes requested):
IMO I will accept the changes for `wilder`. However, for `indent-blankline` and `virtualtext`s, they are **intentionally** _not_ background highlighted as this would affect performance when too many such objects are rendered in the same screenview _(for example, scrolling quickly will cause deferred rendering)_.

https://github.com/ayamir/nvimdots/blob/708c1db3595bb3ff2878d24cea243923c4fd423c/lua/modules/ui/config.lua#L251-L255

CharlesChiuGit:
@Jint-lzxy fair enough, will fix.

CharlesChiuGit:
@ayamir take care!

lua/modules/tools/config.lua
CharlesChiuGit: Ok, will do.

CharlesChiuGit:
@Jint-lzxy PTAL.

Jint-lzxy (approved):
LGTM 🎉 Thank you for your efforts!

PR #384, opened 2022-12-25
perf(event): quicker quit

CharlesChiuGit:
nvim-notify causes hanging on quit, use `silent` to prevent hanging.

### DIFF
--- a/lua/core/event.lua
+++ b/lua/core/event.lua
@@ -74,6 +74,8 @@ function autocmd.load_autocmds()
 			{ "FocusGained", "* checktime" },
 			-- Equalize window dimensions when resizing vim window
 			{ "VimResized", "*", [[tabdo wincmd =]] },
+			-- Silent quit, also quicker
+			{ "VimLeavePre", "*", "silent wall" },
 		},
 		ft = {
 			{ "FileType", "alpha", "set showtabline=0" },
--- a/lua/core/event.lua
+++ b/lua/core/event.lua
@@ -75,7 +75,7 @@ function autocmd.load_autocmds()
 			-- Equalize window dimensions when resizing vim window
 			{ "VimResized", "*", [[tabdo wincmd =]] },
 			-- Silent quit, also quicker
-			{ "VimLeavePre", "*", "silent wall" },
+			{ "VimLeavePre", "*", "silent noh" },
 		},
 		ft = {
 			{ "FileType", "alpha", "set showtabline=0" },


Jint-lzxy:
IMO this will lead to a potential problem. @CharlesChiuGit Can you test this for me?

- First create a new autocmd that is triggered by `VimLeave`, such as:
```vim
:autocmd VimLeave * :sleep 2
```
- Then `:q` to quit vim normally
- If neovim works normally, there will be a 2 second delay before exiting.

If you call _write_ instructions following `silent`, then the `VimLeave` event will be ignored in some cases.

CharlesChiuGit:
> IMO this will lead to a potential problem. @CharlesChiuGit Can you test this for me?
> 
> 
> 
> - First create a new autocmd that is triggered by `VimLeave`, such as:
> 
> ```vim
> 
> :autocmd VimLeave * :sleep 2
> 
> ```
> 
> - Then `:q` to quit vim normally
> 
> - If neovim works normally, there will be a 2 second delay before exiting.
> 
> 
> 
> AFAIK If you call _write_ instructions following `silent`, then the `VimLeave` event will be ignored in some cases.

will test as soon as i got home.

jczhang02:
> IMO this will lead to a potential problem. @CharlesChiuGit Can you test this for me?
> 
>     * First create a new autocmd that is triggered by `VimLeave`, such as:
> 
> 
> ```viml
> :autocmd VimLeave * :sleep 2
> ```
> 
>     * Then `:q` to quit vim normally
> 
>     * If neovim works normally, there will be a 2 second delay before exiting.
> 
> 
> AFAIK If you call _write_ instructions following `silent`, then the `VimLeave` event will be ignored in some cases.

I tested the case and the result is that `neovim` will quit quickly instead of `sleep 2`.

CharlesChiuGit:
@jczhang02 thanks! 
@Jint-lzxy then maybe remove "wall"?

Jint-lzxy:
@CharlesChiuGit Then what command will follow `silent`?
```vim
:sil[ent][!] {command}
```

CharlesChiuGit:
maybe a format command?

Jint-lzxy:
@CharlesChiuGit My question is, why would notify hang on quit? Notify is only a part of the rendering component. If the entire `neovim` instance is asked to be closed, notify should not be the reason to prevent it from closing normally?

CharlesChiuGit:
when one use ":wq" to quit nvim, the format_on_save will pop up a half-rendered notify. this will cause quitting to hanging. if only use ":q", then there will be no hanging.

Jint-lzxy:
Ahh I see. Maybe we can try to modify the `render` functions of notify. I will look into it tomorrow if I have time 👍

CharlesChiuGit:
cool! that would be less "hack" then this solution!

Jint-lzxy:
@CharlesChiuGit I have just tested this problem in detail. The delay occurs only when we `:wq` on a modified file, but I think this may be caused by neovim _itself_ writing files to the disk, not notify. Notify didn't even start its rendering in my test cases.

CharlesChiuGit:
> @CharlesChiuGit I have just tested this problem in detail. The delay occurs only when we `:wq` on a modified file, but I think this may be caused by neovim _itself_ writing files to the disk, not notify. Notify didn't even start its rendering in my test cases.

You're probably right. Or maybe it's just that certain nightly build of neovim having this problem. After I updated neovim, quitting neovim feels snappy again. Closing this PR.

Issue #391, opened 2022-12-30
new opened file from nvim-tree didn't replace dashboard's window


Kotodian:
### Neovim version

NVIM v0.8.1 Build type: Release LuaJIT 2.1.0-beta3 编译者 runner@Mac-1668417329633.local

### Operating system/version

macOS 13.0.1

### Terminal name/version

iTerm2 Build 3.4.18

### Branch info

main (Default/Latest)

### Fetch Preferences

HTTPS (use_ssh = false)

### How to reproduce the issue

1. 进入vi
2. NvimTreeOpen
3. 打开一个文件

### Expected behavior

文件显示完整,如下图所示
<img width="1918" alt="image" src="https://user-images.githubusercontent.com/22030659/210031739-04eb8da5-c6dd-4b92-9603-499422bdb6e3.png">

### Actual behavior

实际情况是只显示了一点,但是第二次打开是能够完整显示。
<img width="1888" alt="image" src="https://user-images.githubusercontent.com/22030659/210031764-57fc67fc-29cd-4ac6-843b-05c037d2c415.png">

CharlesChiuGit:
@Kotodian can u try out if this patch fix your problem?

Kotodian:
> @Kotodian can u try out if this patch fixes your problem?

非常感谢

Kotodian:
> 

Thank you, but this problem still exists, but now it happens occasionally.

ayamir:
It works for me.

Kotodian:
> It works for me.

https://user-images.githubusercontent.com/22030659/210044528-5d12c1d4-274c-4580-9c75-c33aff9d5cb0.mov


我现在的做法是进入到~/.config/nvim文件夹下，切换成新的分支，不清楚这样的操作是否正确。

这么做了以后，第一次是能够打开文件的，但是第二次打开其他文件就是直接打不开了,

CharlesChiuGit:
can u try other terminal emulator?

CharlesChiuGit:
did u install any extra plugins?

ayamir:
> > It works for me.
> 
>  Screen.Recording.2022-12-30.at.15.15.14.mov 
> 我现在的做法是进入到~/.config/nvim文件夹下，切换成新的分支，不清楚这样的操作是否正确。
> 
> 这么做了以后，第一次是能够打开文件的，但是第二次打开其他文件就是直接打不开了,

You have not `PackerSync` because it see nvim only report 89 plugins.

[...]

PR #392, opened 2022-12-30
fix: use `s1n7ax/nvim-window-picker` to provide picker for `nvim-tree`

ayamir:


ayamir:
reference: https://github.com/nvim-tree/nvim-tree.lua/blob/9e4c39572f631bb60ee15cb9d46e1daa9927a45e/doc/nvim-tree-lua.txt#L1009

ayamir:
Add a new plugin can solve #391, but I don't know whether there is a better solution.

CharlesChiuGit:
I can't seem to reproduce #391, even after update all plugins and nvim itself(0.9.0 nightly).

CharlesChiuGit:
I can only confirmed that after using this patch, nvim-tree still works as expected as usual.

CharlesChiuGit (approved):
LGTM.

Jint-lzxy (changes requested):
I don't think this is the problem. I haven't encountered similar issues when using macOS+iTerm2 with [my configuration](https://github.com/Jint-lzxy/nvimconfig). Guess this is related to buffer events.

I'll do a git bisect now to check which commit introduced this problem 👍

Jint-lzxy:
A reliable reproduction steps of the problem is here: https://github.com/ayamir/nvimdots/issues/391#issuecomment-1367827005. Guess this is an upstream issue.

Jint-lzxy (approved):
https://github.com/ayamir/nvimdots/issues/391#issuecomment-1368662179

CharlesChiuGit:
nvim-tree has introduce a new event call `TreeAttachedPost`, maybe we can make use of it.

Issue #395, opened 2023-01-03
Treesitter parsers failed to install


hhcwyx:
### Neovim version

NVIM0.8.2

### Operating system/version

archlinux

### Terminal name/version

alacritty

### Branch info

main (Default/Latest)

### Fetch Preferences

SSH (use_ssh = true)

### How to reproduce the issue

启动nvim,创建一个新文件,点击i准备输入内容时,出现以下信息:
[nvim-treesitter] [0/17] Downloading tree-sitter-css...
nvim-treesitter[lua]: Error during download, please verify your internet connection
请按 ENTER 或其它命令继续
nvim-treesitter[go]: Error during download, please verify your internet connection
请按 ENTER 或其它命令继续
请按 ENTER 或其它命令继续
nvim-treesitter[gomod]: Error during download, please verify your internet connection
请按 ENTER 或其它命令继续
请按 ENTER 或其它命令继续
nvim-treesitter[json]: Error during download, please verify your internet connection
请按 ENTER 或其它命令继续
请按 ENTER 或其它命令继续
nvim-treesitter[latex]: Error during download, please verify your internet connection
请按 ENTER 或其它命令继续
请按 ENTER 或其它命令继续
nvim-treesitter[javascript]: Error during download, please verify your internet connection

请按 ENTER 或其它命令继续
不知道怎么回事?我的网是正常的啊,浏览器访问github也是能访问的啊.

### Expected behavior

怎么才能正常安装,消除以上信息.

CharlesChiuGit:
Does this kind of error only happens when creating new files?

CharlesChiuGit:
Did all the parsers install successfully when you install this repo for the first time?

ayamir:
Have you configured the `ssh` config for accessing github? If not, please use https to update plugins and download treesitter parser.
change `true` to `false` then open file
https://github.com/ayamir/nvimdots/blob/5ccb4b6f2c843b7baa5bcb6f25a4a2088a719917/lua/core/settings.lua#L4

hhcwyx:
> 
设置了ssh,检查了nvimdots/lua/core/settings.lua 第四行没问题.
我用:TSInstall lua 手动试了一下,成功了,但是并不是所有的都成功,比如
TSInstall yaml就没成功.

hhcwyx:
我用:nvim +PackerSync 是正常的.

CharlesChiuGit:
[...]

PR #397, opened 2023-01-03
Change loading order to avoid `cmp not found` when opening new file.

CharlesChiuGit:
and some minor improvements for loading order.
fix part of the issue from #395

### DIFF
--- a/lua/modules/completion/plugins.lua
+++ b/lua/modules/completion/plugins.lua
@@ -27,7 +27,7 @@ completion["glepnir/lspsaga.nvim"] = {
 completion["ray-x/lsp_signature.nvim"] = { opt = true, after = "nvim-lspconfig" }
 completion["hrsh7th/nvim-cmp"] = {
 	config = conf.cmp,
-	event = "InsertEnter",
+	event = "BufReadPost",
 	requires = {
 		{ "onsails/lspkind.nvim" },
 		{ "lukas-reineke/cmp-under-comparator" },
--- a/lua/modules/completion/plugins.lua
+++ b/lua/modules/completion/plugins.lua
@@ -27,11 +27,11 @@ completion["glepnir/lspsaga.nvim"] = {
 completion["ray-x/lsp_signature.nvim"] = { opt = true, after = "nvim-lspconfig" }
 completion["hrsh7th/nvim-cmp"] = {
 	config = conf.cmp,
-	event = "BufReadPost",
+	event = "InsertEnter",
 	requires = {
-		{ "onsails/lspkind.nvim" },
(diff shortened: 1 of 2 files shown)


hhcwyx:
@CharlesChiuGit 按照[b95507d]的指导,修改了相关文件,一切正常了!!!非常感谢!谢谢手把手的精心指导,谢谢!

ayamir (approved):

lua/modules/completion/plugins.lua
```diff
@@ -29,9 +29,9 @@ completion["hrsh7th/nvim-cmp"] = {
 	config = conf.cmp,
 	event = "InsertEnter",
 	requires = {
-		{ "onsails/lspkind.nvim" },
-		{ "lukas-reineke/cmp-under-comparator" },
-		{ "saadparwaiz1/cmp_luasnip", after = "LuaSnip" },
+		{ "onsails/lspkind.nvim", module = "lspkind" },
+		{ "lukas-reineke/cmp-under-comparator", module = "cmp-under-comparator" },
+		{ "saadparwaiz1/cmp_luasnip", after = { "nvim-cmp", "LuaSnip" } },
```
Jint-lzxy, line 34: @CharlesChiuGit A small question here, why `cmp_luasnip` is set both as a dependency of `nvim-cmp` and as being loaded after `nvim-cmp`?

lua/modules/editor/plugins.lua
```diff
@@ -13,6 +13,11 @@ editor["terrortylor/nvim-comment"] = {
 }
 editor["nvim-treesitter/nvim-treesitter"] = {
 	opt = true,
+	module = {
+		"nvim-treesitter",
+		"nvim-treesitter.parsers",
+		"nvim-treesitter.ts_utils",
+	},
```
Jint-lzxy, line 20: Another problem: Directly opening a file by passing in command line parameters after this commit will cause cursor to be unable to move. After removing these lines, the problem disappears.

lua/modules/completion/plugins.lua
CharlesChiuGit: @Jint-lzxy You're correct, there's not need to add "nvim-cmp". I just copy it from my own nvimdot. Thanks for correcting me.

lua/modules/editor/plugins.lua
CharlesChiuGit: > Another problem: Directly opening a file by passing in command line parameters after this commit will cause cursor to be unable to move. After removing these lines, the problem disappears.

Interesting, I don't have this issue in the branch, but thanks for letting me know.

PR #400, opened 2023-01-04
fix(enew): Ensure treesitter is in `rtp`

Jint-lzxy:
This commit ensures that `treesitter` can be loaded correctly and the load-chain will not be corrupted when creating an unnamed buffer _(currently the components of `nvim-treesitter` will not be appended to the runtime path)_. The problem is described in: https://github.com/ayamir/nvimdots/pull/397#discussion_r1061090866.


ayamir (approved):

CharlesChiuGit:
Thanks!

PR #402, opened 2023-01-04
feat(lspsaga & lualine): Show symbols in lualine.

Jint-lzxy:
https://github.com/ayamir/nvimdots/issues/398#issuecomment-1370423444. This can solve rendering problems such as displaying `'cursorcolumn'`. Symbol_in_winbar will cause the space between bufferline and the file to be inaccessible by built-in features.

This is a rough setup and should require additional testing 👍

### DIFF
--- a/lua/modules/utils/init.lua
+++ b/lua/modules/utils/init.lua
@@ -6,6 +6,7 @@ local function hexToRgb(c)
 	return { tonumber(c:sub(2, 3), 16), tonumber(c:sub(4, 5), 16), tonumber(c:sub(6, 7), 16) }
 end
 
+--- Blend foreground with background
 ---@param foreground string @The foreground color
 ---@param background string @The background color to blend with
 ---@param alpha number|string @Number between 0 and 1 for blending amount.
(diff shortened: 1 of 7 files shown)


CharlesChiuGit:
I think a better way to do this is to include the filename in the winbar too, so it will always has a place holder for winbar.
Like this:
圖片

Jint-lzxy:
> I think a better way to do this is to include the filename in the winbar

IMO this is similar to the function of bufferline _(and their location is very close)_. Besides, the problem of built-in rendering has not been resolved.

Jint-lzxy:
I have made adjustments to support `click to jump` 👍 

https://github.com/ayamir/nvimdots/blob/f08d88fca47729931ec319e411d192fb255357de/lua/modules/completion/config.lua#L106-L127

This setting is position independent because `lspsaga` has already processed the formatted string when returning.

CharlesChiuGit:
Also, IMO, the center section of lualine is more suitable for git-blames.

圖片

With this setup, we won't see git-blames floating around all the time which is very distracting.

CharlesChiuGit:
> I have made adjustments to support `click to jump` 👍
> 
> https://github.com/ayamir/nvimdots/blob/f08d88fca47729931ec319e411d192fb255357de/lua/modules/completion/config.lua#L106-L127
> 
> This setting is position independent because `lspsaga` has already processed the formatted string when returning.

Ok, that's pretty cool! Nice!

Jint-lzxy:
> With this setup, we won't see git-blames floating around all the time which is very distracting.

It is rendered as virtual text when you hover over a line :)

CharlesChiuGit:
Yeah, it is virtual text. But it still makes me feel distracting, never mind, it's just my opinion. I'll keep it for myself.

Jint-lzxy:
> Yeah, it is virtual text. But it still makes me feel distracting, never mind, it's just my opinion. I'll keep it for myself.

Another reason for putting symbols here is there are no elements in the center of `lualine` now. Of course, rendering git blame is also a great idea 👍 Let's wait for @ayamir to review these changes :)

CharlesChiuGit:
>  Besides, the problem of built-in rendering has not been resolved.

I don't think it's lspsaga's fault to blame. It's an upstream issue. In neovim nightly(0.9.0), the filename placeholder in winbar will always persist, but in 0.8.2, the rendering issue mentioned in https://github.com/ayamir/nvimdots/issues/398#issuecomment-1370430431 does exist.

圖片

Jint-lzxy:
> I don't think it's lspsaga's fault to blame.

This occurs only when `show_file = false` is set.

The main issue is that currently `lspsaga` will occupy a space between `bufferline` and the file, and this space is independent of file rendering. This means that any built-in full-screen rendering feature _(For example, rendering vertical bars that occupy the entire screen)_ will not touch `bufferline` _(i.e., there will always be a gap between `bufferline` and the file if `lspsaga` doesn't render here)_

lua/modules/utils/init.lua
```diff
@@ -0,0 +1,80 @@
+local M = {}
+
+---@param c string @The color in hexadecimal.
+local function hexToRgb(c)
```
CharlesChiuGit, line 4: Nice function!( just kidding lol

ayamir:
Nice feature! Always showing filename in winbar indeed duplicates with `bufferline`. Tested for click support and it works normally. LGTM 👍

ayamir (approved):

PR #404, opened 2023-01-04
chore(catppuccin): Update highlights

Jint-lzxy:
This commit adjusted some highlights for `catppuccin`, namely:
- The background of floating window (`NormalFloat`) is set to be the same as the main background _(because we added border for all floating window, setting the same color can make color matching more consistent)_;
- The _italic_ attribute of the built-in type of `C/C++` is **disabled** to keep it consistent with the user-defined type;
- _(misc)_ Add border for `lsp_signature`.

**Effect:**
- `lsp_signature`:
lsp-signature  

- Floating window:
normalfloat

### DIFF
--- a/lua/modules/completion/lsp.lua
+++ b/lua/modules/completion/lsp.lua
@@ -37,7 +37,9 @@ local function custom_attach(client, bufnr)
 		fix_pos = true,
 		hint_enable = true,
 		hi_parameter = "Search",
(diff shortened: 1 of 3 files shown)


Jint-lzxy:
Oh I forgot to consider transparent background 😵‍💫

ayamir (approved):

PR #412, opened 2023-01-07
perf(copilot): update lazy loading for copilot

CharlesChiuGit:


Jint-lzxy (commented):
IMO this is unnecessary. No code would require `copilot` before the `VimEnter` event.

CharlesChiuGit:
U r right about that, but I think it's more like a safety measure. Also, the `copilot.lua` recommend adding this.

Jint-lzxy (approved):
I don't use copilot, but LGTM ✨

PR #415, opened 2023-01-10
perf(copilot): disable `panel` & `suggestion` for it can interfere `copilot-cmp`

CharlesChiuGit:
`panel` and `suggestion` are default to true.
Also, the copilot suggestion in cmp seems faster after setting it to false.
ref: https://github.com/zbirenbaum/copilot-cmp/commit/b732a58ac8b7287b981cd9f0d9c0f61e5e9d5760


Jint-lzxy (approved):
LGTM 🍊

PR #436, opened 2023-01-26
fix(nvim-tree): Replace alpha with the target file

### LINKED ISSUE #391: new opened file from nvim-tree didn't replace dashboard's window

Jint-lzxy:
Fixes #391.

The source of the problem is that `alpha` sets `opt_local.buftype = "nofile"`, which will prevent `nvim-tree` from replacing it _(as `nofile` is in the exclusion list)_.  Someone in the `alpha-nvim` repo has once opened an issue _(https://github.com/goolord/alpha-nvim/issues/114)_, but it has never been resolved.
Detail: https://github.com/ayamir/nvimdots/issues/391#issuecomment-1405086281.

Note: We may need to add some filetypes to the list since the buffers of plugins such as `Man.lua` will be recognized, but they should never be replaced IMO.

### DIFF
--- a/lua/core/event.lua
+++ b/lua/core/event.lua
@@ -31,7 +31,7 @@ vim.api.nvim_create_autocmd("BufEnter", {
 
 function autocmd.load_autocmds()
 	local definitions = {
-		packer = {},
+		lazy = {},
 		bufs = {
 			-- Reload vim config automatically
 			{
--- a/lua/modules/completion/lsp.lua
+++ b/lua/modules/completion/lsp.lua
@@ -110,7 +110,7 @@ for _, server in ipairs(mason_lsp.get_installed_servers()) do
 			on_attach = custom_attach,
 			settings = {
 				Lua = {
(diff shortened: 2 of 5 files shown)


CharlesChiuGit:
Can't really review this, since i couldn't reproduce the original bug with this config.

ayamir (approved):

PR #452, opened 2023-01-29
feat(script): Complete installation script on Windows.

### LINKED ISSUE #448: Support for using [un]installation script on Windows
### Feature description

I'm trying to write an installation script for Windows _(`.ps1`)_. In this way, almost all users can have a completely consistent installation experience. This can also eliminate many installation issues IMO.

### Additional information

_No response_

Jint-lzxy:
Resolves #448. @ayamir @CharlesChiuGit Can you test this script on Windows? It executes correctly on macOS running `pwsh core`.

Command:
```pwsh
Invoke-WebRequest "https://raw.githubusercontent.com/Jint-lzxy/nvimdots/feat/windows-installer/install/install.ps1" -UseBasicParsing | Invoke-Expression
```

### DIFF
--- a/install/install.ps1
+++ b/install/install.ps1
@@ -110,7 +110,7 @@ function Check-Def-Exe ([Parameter(Mandatory = $True,ValueFromPipeline = $True)]
 
 function Query-Pack {
 	if ((Check-Def-Exe -WithName "scoop") -and (Check-Def-Exe -WithName "choco")) {
-		prompt -Msg " [Detected] Multiple package mgrs detected."
+		prompt -Msg "   [Detected] Multiple package mgrs detected."
 
 		$_title = "Package manager Preferences"
 		$_message = "Pick your favorite package manager"
@@ -125,10 +125,10 @@ function Query-Pack {
 			$env:CCPACK_MGR = 'choco'
 		}
 	} elseif ((Check-Def-Exe -WithName "scoop")) {
-		prompt -Msg " [Detected] We'll use 'Scoop' as the default package mgr."
+		prompt -Msg "   [Detected] We'll use 'Scoop' as the default package mgr."
 		$env:CCPACK_MGR = 'scoop'
 	} elseif ((Check-Def-Exe -WithName "choco")) {
-		prompt -Msg " [Detected] We'll use 'Chocolatey' as the default package mgr."
+		prompt -Msg "   [Detected] We'll use 'Chocolatey' as the default package mgr."
 		$env:CCPACK_MGR = 'choco'
 	} else {
 		_abort -Msg "Required executable not found." -Type "NotInstalled" -Info_msg @'
--- a/install/install.ps1
+++ b/install/install.ps1
@@ -221,7 +221,7 @@ Safe-Execute -WithCmd { Set-Location -Path "$env:CCDEST_DIR" }
 
 if (-not $USE_SSH) {
 	prompt -Msg "Changing default fetching method to HTTPS..."
-	Safe-Execute -WithCmd { (Get-Content "$CCDEST_DIR\lua\core\settings.lua") | ForEach-Object { $_ -replace '\[\"use_ssh\"\] \= true','\[\"use_ssh\"\] \= false' } | Set-Content "$CCDEST_DIR\lua\core\settings.lua" -WhatIf }
+	Safe-Execute -WithCmd { (Get-Content "$env:CCDEST_DIR\lua\core\settings.lua") | ForEach-Object { $_ -replace '\[\"use_ssh\"\] \= true','\[\"use_ssh\"\] \= false' } | Set-Content "$env:CCDEST_DIR\lua\core\settings.lua" -WhatIf }
 }
 
 prompt "Spawning neovim and fetching plugins... (You'll be redirected shortly)"
--- a/install/install.ps1
+++ b/install/install.ps1
@@ -184,11 +184,13 @@ function _install_nodejs_deps {
 }
 
 function _install_python_deps {
+	Safe-Execute -WithCmd { python -m pip install --user wheel }
 	Safe-Execute -WithCmd { python -m pip install --user pynvim }
 }
 
 function _install_ruby_deps {
 	Safe-Execute -WithCmd { gem install neovim }
+	Safe-Execute -WithCmd { ridk install }
 }
 
 function Check-And-Fetch-Exec ([Parameter(Mandatory = $True,ValueFromPipeline = $True)][ValidateNotNullOrEmpty()] [string]$PkgName) {
@@ -351,8 +353,8 @@ if (-not $USE_SSH) {
 	}
 }
 
-prompt "Spawning neovim and fetching plugins... (You'll be redirected shortly)"
-prompt 'If lazy.nvim failed to fetch any plugin(s), maunally execute `:Lazy sync` until everything is up-to-date.'
+prompt -Msg "Spawning neovim and fetching plugins... (You'll be redirected shortly)"
+prompt -Msg 'If lazy.nvim failed to fetch any plugin(s), maunally execute `:Lazy sync` until everything is up-to-date.'
 Write-Host @'
 
 Thank you for using this set of configuration!
--- a/install/install.ps1
+++ b/install/install.ps1
@@ -41,8 +41,9 @@ function Test-OpType {
 	$NonInteractive = [Environment]::GetCommandLineArgs() | Where-Object { $_ -like '-NonI*' }
 	if ([Environment]::UserInteractive -and -not $NonInteractive) {
 		return $False
+	} else {
+		return $True
 	}
-	return $True
 }
 
 function prompt ([Parameter(Mandatory = $True,ValueFromPipeline = $True)][ValidateNotNullOrEmpty()] [string]$Msg) {
@@ -65,7 +66,6 @@ function Safe-Execute ([Parameter(Mandatory = $True,ValueFromPipeline = $True)][
 	}
 	catch {
 		_abort -Msg "Failed during: $WithCmd" -Type "InvalidResult"
-		exit 1
 	}
 }
 
@@ -156,7 +156,7 @@ function Init-Pack {
 		} elseif (($env:CCPACK_MGR -eq 'scoop') -and (Check-Def-Exe -WithName $env:CCPACK_MGR)) {
 			prompt -Msg "We'll use 'Scoop' as the default package mgr."
 		} else {
-			prompt -Msg "Validation faled. Fallback to query."
+			prompt -Msg "Validation failed. Fallback to query."
 			Query-Pack
 		}
 	} else {
@@ -191,7 +191,7 @@ function _install_ruby_deps {
 	Safe-Execute -WithCmd { gem install neovim }
 }
 
-function Check-And-Fetch-Exe ([Parameter(Mandatory = $True,ValueFromPipeline = $True)][ValidateNotNullOrEmpty()] [string]$PkgName) {
+function Check-And-Fetch-Exec ([Parameter(Mandatory = $True,ValueFromPipeline = $True)][ValidateNotNullOrEmpty()] [string]$PkgName) {
 	if (-not (Check-Def-Exe -WithName $PkgName)) {
 		Write-Host "Checking dependency: '$PkgName'`t" -NoNewline
 		Start-Sleep -Milliseconds 350
@@ -226,18 +226,18 @@ function Check-Dep-Choice ([Parameter(Mandatory = $True,ValueFromPipeline = $Tru
 }
 
 function Fetch-Deps {
-	Check-And-Fetch-Exe -PkgName "git"
-	Check-And-Fetch-Exe -PkgName "nvim"
-	Check-And-Fetch-Exe -PkgName "make"
-	Check-And-Fetch-Exe -PkgName "node"
-	Check-And-Fetch-Exe -PkgName "pip"
-	Check-And-Fetch-Exe -PkgName "fzf"
-	Check-And-Fetch-Exe -PkgName "ruby"
-	Check-And-Fetch-Exe -PkgName "go"
-	Check-And-Fetch-Exe -PkgName "curl"
-	Check-And-Fetch-Exe -PkgName "wget"
-	Check-And-Fetch-Exe -PkgName "rustc"
-	Check-And-Fetch-Exe -PkgName "tree-sitter"
+	Check-And-Fetch-Exec -PkgName "git"
+	Check-And-Fetch-Exec -PkgName "nvim"
+	Check-And-Fetch-Exec -PkgName "make"
+	Check-And-Fetch-Exec -PkgName "node"
+	Check-And-Fetch-Exec -PkgName "pip"
+	Check-And-Fetch-Exec -PkgName "fzf"
+	Check-And-Fetch-Exec -PkgName "ruby"
+	Check-And-Fetch-Exec -PkgName "go"
+	Check-And-Fetch-Exec -PkgName "curl"
+	Check-And-Fetch-Exec -PkgName "wget"
+	Check-And-Fetch-Exec -PkgName "rustc"
+	Check-And-Fetch-Exec -PkgName "tree-sitter"
 }
 
 function Is-Latest {
@@ -278,7 +278,7 @@ if ((Check-Dep-Choice -PkgName "Ruby")) {
 }
 
 # Check dependencies
-if ((Get-Command "nvim" -ErrorAction SilentlyContinue) -eq $null) {
+if (-not (Check-Def-Exe -WithName "nvim")) {
 	_abort -Msg "Required executable not found." -Type "NotInstalled" -Info_msg @'
 You must install NeoVim before installing this Nvim config. See:
   https://github.com/neovim/neovim/wiki/Installing-Neovim
@@ -288,7 +288,7 @@ You must install NeoVim before installing this Nvim config. See:
 '@
 }
 
-if ((Get-Command "git" -ErrorAction SilentlyContinue) -eq $null) {
+if (-not (Check-Def-Exe -WithName "git")) {
 	_abort -Msg "Required executable not found." -Type "NotInstalled" -Info_msg @'
 You must install Git before installing this Nvim config. See:
   https://git-scm.com/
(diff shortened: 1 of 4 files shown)


Jint-lzxy:
It installs the configuration to `$env:LOCALAPPDATA\nvim` by default.

CharlesChiuGit:
I just ran the command directly. Do I need to set sth else?

圖片

ayamir:
Execute this command directly failed. 

image


But it succeeded after I download this script manually and execute it.

CharlesChiuGit:
I ran the script on a fresh win11 machine with following steps:

1. ```Set-ExecutionPolicy -Scope Process -ExecutionPolicy Bypass``` to make the ps1 file can be runable.
2. Install scoop and `scoop install neovim`.
3. Grep the script and run it.
4. Most of the plugin cloned successfully, but the plugins require other tools like `nodejs`, `cargo`, `make`, `cmake`, `python` etc will fail as expected.

圖片

```log
[ERROR 2023/1/30 ¤W¤È 04:01:01] ...a\site\lazy\mason.nvim/lua/mason-core/installer/init.lua:159: Installation failed for Package(name=black) error="...te\\lazy\\mason.nvim/lua/mason-core/managers/pip3/init.lua:90: Unable to create python3 venv environment."
[ERROR 2023/1/30 ¤W¤È 04:01:01] ...a\site\lazy\mason.nvim/lua/mason-core/installer/init.lua:159: Installation failed for Package(name=prettier) error=spawn: npm.cmd failed with exit code - and signal -. npm.cmd is not executable
```

Things that failed to install or having errors:
1. tree-sitter parsers
2. Luasnip(fix: install `make`)
3. telescoop-fzf-native.nvim(fix: `scop install fzf)
4. markdown-preview.nvim(fix: install `nodejs`)
5. sniprun(fix: install `cargo`)
6. prettier(fix: install `nodejs`)
7. black(fix: install `python`?)
8. pyright
9. bashls
10. gopls
11. efm

I think the script is good to go.

Just the command you mentioned is still not able to run directly with following error. (Powershell 7.3.2)
圖片

Jint-lzxy:
Thanks for the information 👍 I will try to use `Winget`/`Chocolatey`/`Scoop` to fetch necessary packages _(Users may choose from them)_.

Jint-lzxy:
> Execute this command directly failed.
> 
> image
> 
> But it succeeded after I download this script manually and execute it.

All necessary environment variables will be set before this in the next push 👍

Jint-lzxy:
So now we need to:
1. Properly setup `pwsh` to accept scripts from any source;
2. Set up the environment correctly _(this script cannot run in a completely clean environment as `Set-StrictMode -Version 3.0`)_
3. Complete the installation of dependencies in the install script

Jint-lzxy:
> > I just ran the command directly. Do I need to set sth else? btw, I'm using `powersehll`, not `windows powershell`. Does that make a difference?
> > 圖片
> 
> It just require `powershell` version `7.1`, it's `powershell` instead of `windows powershell`.

Emm This script would only need `pwsh core` to execute successfully. It doesn't matter what pre-/suffixes `pwsh` have... I will fix these errors on my Windows 10 VM later.

Jint-lzxy:
> I just ran the command directly. Do I need to set sth else? btw, I'm using `powersehll`, not `windows powershell`. Does that make a difference?
> 
> 圖片

Maybe this issue is related to the prefix `???`? I will use raw strings or just download it to the `TMP` folder and then run the script there.

Jint-lzxy:
`(System.Net.WebClient).DownloadString` will be used instead, this works fine

CharlesChiuGit:
> Thanks for the information 👍 I will try to use `Winget`/`Chocolatey`/`Scoop` to fetch necessary packages _(Users may choose from them)_.

`winget` would be nice! Since it's built-in, but I'm not sure if winget can handle github source that well, like `fzf` or most of the lsps.

Jint-lzxy:
> > Thanks for the information 👍 I will try to use `Winget`/`Chocolatey`/`Scoop` to fetch necessary packages _(Users may choose from them)_.
> 
> `winget` would be nice! Since it's built-in, but I'm not sure if winget can handle github-sourced tools that well, like `fzf` or most of the lsps.

Hmm then IMO we should just use `Chocolatey` or `Scoop`. Although we have `Mason` in nvim, some of its dependencies still can't be installed by `winget` :(

I think we can support `winget` after it supports all packages we need.

CharlesChiuGit:
Yeah, totally agreed. Personally I like `scoop` better since it doesn't need admin privileges, which would be vital for ppl who wants to install this config on their work machine.

Jint-lzxy:
@CharlesChiuGit I have a small question. Now the script will try to use a command and its `.exe` suffix form _(e.g., `git` and `git.exe`)_ to avoid the case when users didn't add the symlink of non-suffixal form _(e.g., `git`)_ to his `PATH`.
```pwsh
if ($hasEXEnvim) {
  $nvim_version = Invoke-Command -ErrorAction SilentlyContinue -ScriptBlock { **nvim.exe** --version }
} else {
  $nvim_version = Invoke-Command -ErrorAction SilentlyContinue -ScriptBlock { **nvim** --version }
}
```
Are those checks necessary? I'm not a Windows user, so don't know much about this :(

Jint-lzxy:
I decided to remove these checks. It should be the user's responsibility to ensure that the command exists.

CharlesChiuGit:
I think it's not necessary, since we're running the command in a script. For example I installed `git.exe`, but I can call `git` in `pwsh` as in linux/macOS.(Sorry I got caught up by my boss.

圖片

Jint-lzxy:
> > I just ran the command directly. Do I need to set sth else? btw, I'm using `powersehll`, not `windows powershell`. Does that make a difference?
> > 圖片
> 
> Maybe this issue is related to the prefix `???`? I will use raw strings or just download it to the `TMP` folder and then run the script there.

Update: Located this issue:
```console
❯ xxd install/install.ps1
00000000: efbb bf23 5265 7175 6972 6573 202d 5665  ...#Requires -Ve
```
🤦‍♂️🤦‍♂️There is an issue with the formatter as it runs on Windows by default, should tell we're running on macOS :(

Jint-lzxy:
https://github.com/ayamir/nvimdots/pull/452/commits/7a58e6fd6500c88322ee0d5a93ab7e1c1e3a14af implemented package manager support.

The correct command is:
```pwsh
Set-ExecutionPolicy Bypass -Scope Process -Force; Invoke-Expression ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/Jint-lzxy/nvimdots/feat/windows-installer/install/install.ps1'))
```

Jint-lzxy:
@CharlesChiuGit @ayamir Now all the above issues have been resolved 👍

When testing, it is best to uninstall (some of) the following dependencies to test whether `Check-And-Fetch-Exe` works properly
```pwsh
$choco_package_matrix = @{ "git" = "git"; "nvim" = "neovim"; "make" = "make"; "node" = "nodejs"; "pip" = "python3"; "fzf" = "fzf"; "go" = "go"; "curl" = "curl"; "wget" = "wget"; "tree-sitter" = "tree-sitter"; "ruby" = "ruby"; "rustc" = "rust-ms" }
```

CharlesChiuGit:
Will test after I got home!

CharlesChiuGit:
圖片
圖片

I stuck at the second picture, running with scoop.

Jint-lzxy:
Fixed at https://github.com/ayamir/nvimdots/pull/452/commits/c35905236045a17bc78c13db0464b9f2a58f434d. Also run `npm install -g npm@9.4.0` first 👍

Jint-lzxy:
Wait a minute. The real issue should be `pip` is not initialized.

CharlesChiuGit:
I found one interesting issue. It seems like the newly installed app aren't being registered in PATH unless one do a logout/login or a restart. `node` and `ruby` didn't being pick up after fresh install, even open a new shell.

Jint-lzxy:
> I found one interesting issue. It seems like the newly installed app aren't being registered in PATH unless one do a logout/login or a restart. `node` and `ruby` didn't being pick up after fresh install, even open a new shell.

@CharlesChiuGit Yeah I also realized that. Will do:
```pwsh
$env:Path = [System.Environment]::GetEnvironmentVariable("Path","Machine") + ";" + [System.Environment]::GetEnvironmentVariable("Path","User") 
```

Jint-lzxy:
Fixed at https://github.com/ayamir/nvimdots/pull/452/commits/5974f70f029d80a86a4bfb2a6096aae4ea90103c

Jint-lzxy:
> I found there's a `dark` package being installed before installing `python`, is that intended?

Yeah I believe that's intended 😄AFAIK `dark` is the WiX (Windows Installer XML) Toolset Decompiler, to install `python` we would need that decompiler

CharlesChiuGit:
圖片

I'm not sure if MSYS2 is needed, but it seems nice to include a line of `ridk install`?

圖片

I think we should consider including `wheel` and `pyproject.toml` before installing `pynvim`, for it might be deprecated soon.

圖片

And finally, I still got that weird error below after all plugins and ts parsers are install successfully by lazy.

```ps1
cmdlet prompt at command pipeline position 1
Supply values for the following parameters:
Msg:
```

CharlesChiuGit:
One small thing:
I think we should consider replacing this line
https://github.com/ayamir/nvimdots/blob/c313d5262a9b224e2958b1c31bd7097473c69861/lua/modules/lang/plugins.lua#L19

with 
```lua
	build = function()
		vim.fn["mkdp#util#install"]()
	end,
```
The later seems work faster than the previous one.

Jint-lzxy:
Fixed at: https://github.com/ayamir/nvimdots/pull/452/commits/a8bf97bf7f73032c0aa6fd27a722c8b57327ceda

Jint-lzxy:
> One small thing: I think we should consider replacing this line
> 
> https://github.com/ayamir/nvimdots/blob/c313d5262a9b224e2958b1c31bd7097473c69861/lua/modules/lang/plugins.lua#L19
> 
> with
> 
> ```lua
> 	build = function()
> 		vim.fn["mkdp#util#install"]()
> 	end,
> ```
> 
> The later seems work faster than the previous one. (I use the later in my personal config and it worked fine.

Good catch! Just a quick note, the vim function can be directly called via:
```lua
lang["iamcco/markdown-preview.nvim"] = {
	lazy = true,
	ft = "markdown",
	-> build = ":call mkdp#util#install()",
}
```

CharlesChiuGit:
Interesting! Will steal. lol

install/install.ps1
```diff
@@ -0,0 +1,372 @@
...
+
+prompt -Msg "Fetching in progress..."
+
+if ($USE_SSH) {
+	if ((Is-Latest)) {
+		Safe-Execute -WithCmd { git clone -b "$env:CCLONE_BRANCH" 'git@github.com:ayamir/nvimdots.git' "$env:CCDEST_DIR" }
+	} else {
+		warn -Msg "You have outdated Nvim installed (< $REQUIRED_NVIM_VERSION)."
+		prompt -Msg "Automatically redirecting you to legacy version..."
+		Safe-Execute -WithCmd { git clone -b 0.7 'git@github.com:ayamir/nvimdots.git' "$env:CCDEST_DIR" }
```
CharlesChiuGit, line 333: I think we should use `git clone --depth=1` to avoid cloning extra things to the user.

install/install.ps1
```diff
@@ -0,0 +1,372 @@
...
+	}
+} else {
+	if ((Is-Latest)) {
+		Safe-Execute -WithCmd { git clone -b "$env:CCLONE_BRANCH" 'https://github.com/ayamir/nvimdots.git' "$env:CCDEST_DIR" }
```
CharlesChiuGit, line 337: Also this line too.

install/install.ps1
```diff
@@ -0,0 +1,372 @@
...
+	} else {
+		warn -Msg "You have outdated Nvim installed (< $REQUIRED_NVIM_VERSION)."
+		prompt -Msg "Automatically redirecting you to legacy version..."
+		Safe-Execute -WithCmd { git clone -b 0.7 'https://github.com/ayamir/nvimdots.git' "$env:CCDEST_DIR" }
```
CharlesChiuGit, line 341: And this line lol

CharlesChiuGit (changes requested):
Minor thingy

CharlesChiuGit:
圖片

Still the same QQ

CharlesChiuGit:
> 圖片
> 
> Still the same QQ

I found this error occurs only when using 
```ps1
Invoke-Expression ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/Jint-lzxy/nvimdots/feat/windows-installer/install/install.ps1'))
```

If I download the script and run it, the script will end successfully.

CharlesChiuGit:
I tried to find if there's any missing `-Msg` in the script, but it seems every `_abort`, `prompt`, `warn`, `warn-Ext` all have `-Msg`.

However, I also tried 
```ps1
irm https://raw.githubusercontent.com/Jint-lzxy/nvimdots/feat/windows-installer/install/install.ps1 | iex
```

still having 
```
cmdlet prompt at command pipeline position 1
Supply values for the following parameters:
Msg:
```

So I think it's more likely that sth is wrong in the script.

Jint-lzxy:
> cmdlet prompt at command pipeline position 1
> Supply values for the following parameters:
> Msg:

Will do a debug traceback now 👍

Jint-lzxy:
https://github.com/ayamir/nvimdots/pull/452/commits/98e32aea44d86e3e6f5e7049a10082652fb671f3 added an option for both scripts to check whether the user prefers shallow clone or a complete clone.

Jint-lzxy:
All problems have been resolved. Only one last full test is required before merging 👍

CharlesChiuGit:
Checking now( a lot of pending downloads lol

CharlesChiuGit:
圖片

New shallow clone function seems having a small issue.

Jint-lzxy:
Fixed at https://github.com/ayamir/nvimdots/pull/452/commits/7c575e0e3b619deb54e2c20ce97812eb0ba711a4

CharlesChiuGit:
Why does node pkg always took FOREVER to install... just saying...

Jint-lzxy:
> Why does node pkgs always took FOREVER to install... just saying...

There could be many causes for this... The versions of `npm` and `node` do not match those described in the spec; The user is behind a proxy...

Just answer `no` to NodeJS dependency installation 🥲

CharlesChiuGit:
圖片

I think we need to install `gcc` for this.

Jint-lzxy:
> I think we need to install `gcc` for this and ts parsers. (`scoop install gcc`)

I would choose `LLVM` as it's available on both package managers. (Sadly `msvc` and `gcc` are both only available on one of them ;(

The `LLVM` tool chain also provides some executables required for this configuration 👍

CharlesChiuGit:
圖片

https://tenor.com/vfGo.gif

Jint-lzxy:
🤦‍♂️ @CharlesChiuGit What do you see if u supply that value (;

CharlesChiuGit:
圖片

Then I normally type `exit` in **PS>** and it close the shell.

Jint-lzxy:
🤦‍♂️ OK I'll just remove that `Mandatory = $True` restriction

CharlesChiuGit:
圖片

`llvm` seems to work only partially.

Jint-lzxy:
Then we would still need `msvc` installed.

I would remove `llvm` and let the user install [visual studio](https://visualstudio.microsoft.com/zh-hans/#vs-section) himself.

CharlesChiuGit:
bloody hell msvc 👎

Jint-lzxy:
Make sure to tick `Desktop development with C++` at the very least

Jint-lzxy:
https://github.com/ayamir/nvimdots/pull/452/commits/92c62a0cc2fe1e3bd9a10507d54aabd8fe154856 removed those restrictions 😅 Should work now

CharlesChiuGit:
The result is not ideal but acceptable.

My original pwsh promt:
圖片

pwsh promt after installation:
圖片

I come up with another workaround. Maybe we can open a new terminal window and force quit that window so no one would saw that issue. How do u think?

Jint-lzxy:
(I just checked [mingw](https://scoop.sh/#/apps?q=mingw&s=0&d=1&o=true), it works fine) I personally don't want users to install other editors when installing this **neovim** config 😅

CharlesChiuGit:
> (I just checked [mingw](https://scoop.sh/#/apps?q=mingw&s=0&d=1&o=true), it works fine) I personally don't want users to install other editors when installing this **neovim** config 😅

Lamo 💯 💯 💯 1000% agreed.

Jint-lzxy:
> The result is not ideal but acceptable.
> 
> My original pwsh promt: 圖片
> 
> pwsh promt after installation: 圖片
> 
> I come up with another workaround. Maybe we can open a new terminal window and force quit that window so no one would saw that issue. How do u think? ( I don't there's anyone run Windows headlessly.)
> 
> Aside from that, I think everything is good to go.(I use scoop's gcc for C compiling tho)

Oh God I forgot to exit from the install script :( That's the cause of all the issues
A "bash-sequela" 👎

CharlesChiuGit:
No... I think it that pwsh is just so fking weird. TBO, I'm respect u a lot, since I personally would not want to learn pwsh lol

Jint-lzxy:
Finally https://github.com/ayamir/nvimdots/pull/452/commits/d82f71d619b6f855d4622792780607a0c5658b3f

Jint-lzxy:
@CharlesChiuGit Hope this is the last test **ever** :(

Yeah `pwsh`'s syntax and **behavior** is just so weird.

CharlesChiuGit:
Nice! Everything works fine!

CharlesChiuGit (approved):
LGTM

CharlesChiuGit:
Final command:
```ps1
Set-ExecutionPolicy Bypass -Scope Process -Force; irm https://raw.githubusercontent.com/ayamir/nvimdots/main/install/install.ps1 | iex

**Comment by Jint-lzxy** [comment]
I would prefer full command name and `HEAD` as git ref:
```pwsh
Set-ExecutionPolicy Bypass -Scope Process -Force; Invoke-WebRequest 'https://raw.githubusercontent.com/ayamir/nvimdots/HEAD/install/install.ps1' | Invoke-Expression
```

CharlesChiuGit:
圖片

This guy asked too early lol

PR #458, opened 2023-01-31
refactor(structure): structurize nvim config

CharlesChiuGit:
Firstly, sorry about forgot to create a new branch.
I structured the config in my own sense, @ayamir @Jint-lzxy please let me know if I need to change anything!


CharlesChiuGit:
Some files are being capitalized are due the module conflicting. Wellcome to suggest a better strategy!

Jint-lzxy:
Will push some changes later 👍

Jint-lzxy:
> Some files are being capitalized are due the module conflicting. Wellcome to suggest a better strategy!

It doesn't matter _(also this would have no effect on case-insensitive systems like macOS)_. We can ensure that the relative path is unique, and the user config is always in the first place 👍

CharlesChiuGit:
But the lsp will show warning? Or it's safe to ignore?

Jint-lzxy:
> But the lsp will show warning? Or it's safe to ignore?

Because it assumes that the current directory is also part of `rtp`. But yes it would be safer to have a different file name 👍

Jint-lzxy:
IMO this structure would be cleaner and could avoid the warnings in https://github.com/ayamir/nvimdots/pull/458#issuecomment-1410374189.
```
lua/
└── modules/
    ├── plugins/ <-- Set `lazy.nvim`'s load dir to this **string** and it will automatically load all specs
    │   ├── completion.lua
    │   ├── editor.lua
    │   ├── lang.lua
    │   ├── tools.lua
    │   └── ui.lua
    ├── configs/
    │   ├── completion/
    │   ├── editor/
    │   ├── lang/
    │   ├── tools/
    │   └── ui/
    └── utils/
        ├── icons.lua
        └── init.lua
```

any ideas?

CharlesChiuGit:
So the config files in `configs/completion/`, for example, would be the exact same name as the actual module name? Like `configs/completion/cmp.lua` would be auto-loaded by lazy.nvim?

Jint-lzxy:
> So the config files in `configs/completion/`, for example, would be the exact same name as the actual module name? Like `configs/completion/cmp.lua` would be auto-loaded by lazy.nvim?

Since we've required that file in the spec, then it would behave like `lazy.nvim` auto-loaded this config to memory.

CharlesChiuGit:
If this would work, then I think it's indeed a more sensible structure than my original one.

Jint-lzxy:
```lua
completion["hrsh7th/nvim-cmp"] = {
	event = "InsertEnter",
	config = require("modules.configs.completion.cmp"), <- This line loads the config
	dependencies = {
		.....
	}
}
```
As this file (`plugins/completion.lua`) would be automatically required by `lazy.nvim` then "it would behave like `lazy.nvim` auto-loaded this config to memory."

ayamir:
> IMO this structure would be cleaner and could avoid the warnings in https://github.com/ayamir/nvimdots/pull/458#issuecomment-1410374189.

Agree.

CharlesChiuGit:
I'm still a bit confuse tho. If we still need to specify `config = require("modules.configs.completion.cmp")`, then what does "auto-loaded" means? Or I misunderstand sth?

Jint-lzxy:
> I'm still a bit confuse tho. If we still need to specify `config = require("modules.configs.completion.cmp")`, then what does "auto-loaded" means? Or I misunderstand sth?

Actually, this is not auto-load. It _behaves_ like auto-load. "Auto-load" here means:
```console
Some users may want to split their plugin specs in multiple files.
Instead of passing a spec table to setup(), you can use a Lua module.

The specs from the module and any top-level sub-modules will be
merged together in the final spec, so it is not needed to add require
calls in your main plugin file to the other files.
```

As "**The specs from the module and any top-level sub-modules will be merged together in the final spec**", this auto-merging process _(i.e., requiring all files inside `modules.plugins`)_ would also **require** the _requires_ inside each plugin spec (\*.lua). Then, all configs would be auto-loaded to memory since we've required it's parent `plugins/<spec-name>.lua`.

Here we supply a **string** to the setup function, not a **table**:
```lua
require("lazy").setup("modules.plugins")
```

Jint-lzxy:
Auto-load here means we don't need to fetch and merge all plugin specs before `lazy.nvim` _(specifically all `plugin.lua`s)_. It'll handle this for us.

CharlesChiuGit:
Aha, got it. So it's more like a "auto-merge"?

That would be great! When we were using `packer.nvim`, we need to merge the tables by ourselves. But now `lazy.nvim` will handle this for us, nice!

Jint-lzxy:
> Aha, got it. So it's more like a "auto-merge"?
> 
> That would be great! When we were using `packer.nvim`, we need to merge the tables by ourselves. But now `lazy.nvim` will handle this for us, nice!

Yup that's it ;)

CharlesChiuGit:
For user specific plugins, I think it should be like this.

```txt
lua/
└── modules/
    ├── plugins/
    │   ├── user.lua <--- user plugins
    │   ├── completion.lua
    │   ├── editor.lua
    │   ├── lang.lua
    │   ├── tools.lua
    │   └── ui.lua
    ├── configs/
    │   ├── user/ <--- user plugins' config files
    │   ├── completion/
    │   ├── editor/
    │   ├── lang/
    │   ├── tools/
    │   └── ui/
    └── utils/
        ├── icons.lua
        └── init.lua
```

CharlesChiuGit:
Also, one more question. I'm always curious about why we make extra function when setting keymaps?
Isn't that build up the difficultly for users to modify their own keymaps? 

Why not just do sth like this?
https://github.com/CharlesChiuGit/nvimdots.lua/blob/main/lua/keymap/init.lua

I think it's more native to neovim?

Jint-lzxy:
To be short, it allows users to set \[mode|keybinding\] on one side, and use `lua`'s colon syntax to chain functions on the other side.

Long answer:
```lua
        ┌─ sep     ┌── map_type (saving a lot of keystrokes)
     ["n|gb"] = map_cr("BufferLinePick"):with_noremap():with_silent(),
       │  └── map_key          │              └── special1    │
       └──── map_mode          └── map_content                └── special2 (chained)
```
This should be more intuitive one you get used to this.

CharlesChiuGit:
Thanks for your answer! It's probably looks weird to me due to I only have little experience using C++.

Jint-lzxy:
@CharlesChiuGit @ayamir IMO it would be better to include all pre-configured servers to `mason`'s ensure-installed list to avoid meaningless issues.

ayamir:
> @CharlesChiuGit @ayamir IMO it would be better to include all pre-configured servers to `mason`'s ensure-installed list to avoid meaningless issues.

But I even want to reduce `ensure_installed` list.

CharlesChiuGit:
> @CharlesChiuGit @ayamir IMO it would be better to include all pre-configured servers to `mason`'s ensure-installed list to avoid meaningless issues.

Personally I don't think that's a good idea, it will make the installation time of this config to be super long. I think current `lsp.lua` and `server-configs/` is a good demo already.

CharlesChiuGit:
Also @Jint-lzxy your last few pushes of this draft seems to break a lot of things and I have no time to debug it right now. It'll be nice if some patch fix would be applied soon.

Jint-lzxy:
> > @CharlesChiuGit @ayamir IMO it would be better to include all pre-configured servers to `mason`'s ensure-installed list to avoid meaningless issues.
> 
> Personally I don't think that's a good idea, it will make the installation time of this config to be super long. I think current `lsp.lua` and `server-configs/` is a good demo already.

Agreed. Will do a cleanup after migration.

Jint-lzxy:
> Also @Jint-lzxy your last few pushes of this draft seems to break a lot of things and I have no time to debug it right now. It'll be nice if some patch fix would be applied soon, tks!

Yep I've just finished implementing the file structure described in https://github.com/ayamir/nvimdots/pull/458#issuecomment-1410451718. Will push some patches soon!

Structure Preview: [Jint-lzxy/nvimconfig/tree/experimental/split-configs](https://github.com/Jint-lzxy/nvimconfig/tree/experimental/split-configs).

Jint-lzxy:
Will also add `modules_dir .. "/configs` to the runtime path to simplify function calls

CharlesChiuGit:
> Structure Preview: [Jint-lzxy/nvimconfig/tree/experimental/split-configs](https://github.com/Jint-lzxy/nvimconfig/tree/experimental/split-configs/lua?rgh-link-date=2023-02-01T04%3A04%3A32Z). ([Jint-lzxy/nvimconfig@d9e987f](https://github.com/Jint-lzxy/nvimconfig/commit/d9e987fdc4854534367dae575970511253fb85a8))

Looks awesome! Nice work!

Jint-lzxy:
https://github.com/ayamir/nvimdots/pull/458/commits/948c2d4e6c10add70ae7407aa50cde3d038a8e23 appended lua's native `package.path` _(not vim's `rtp`)_ to save a few keystrokes:
```lua
require("modules.configs.completion.mason-installer") -> require("completion.mason-installer")
```
Also fixes startup issue. Now `nvim` would start without error.

Jint-lzxy:
https://github.com/ayamir/nvimdots/pull/458/commits/f0c5f4e447f72158f2d82c572c7088fed424117b finalized the file structure and adjusted naming. As we are only modifying `lua`'s native search path _(instead of vim's rtp)_, `require()` will only search for files in the `config` directory _(it will never touch plugins' dirs)_. Hence the config files are named as intuitively as possible.

ayamir (approved):
No problems, LGTM 👍

CharlesChiuGit:
Looks fine on Win11. Well do some in-depth tests on both windows and linux when I got home later the night.

Jint-lzxy:
IMO We should also update the wiki before merging. The use of plugins has changed A LOT ;;

ayamir:
IMO we also need to make the boundary of different kinds clearer and make some necessary change to current plugins' kind.

CharlesChiuGit:
> different kinds

maybe **different categories** would be a better word. I struggle a bit to understand your meaning lol. No big deal actually.

CharlesChiuGit:
> IMO we also need to make the boundary of different kinds clearer and make some necessary change to current plugins' kind. So maybe a new classification is needed. It no just make plugins' strcture clearer but has benefit to plugins load sequency and dependency.

I agreed. TBO, I have some opinions on certain plugins in it's current category.

CharlesChiuGit:
> Looks fine on Win11 with some simple tests, like lsp and cmp. Well do some in-depth tests on both windows and linux when I got home later the night.

Sorry I got OT yesterday. Will do it now.

CharlesChiuGit:
圖片

On windows, I will still get this error msg after `scoop install sqlite`.

圖片

But after some searching, I think this issue is unsolvable. Maybe just note it in the wiki.

CharlesChiuGit:
圖片

Also `lualine`'s `cwd` doesn't show on windows.

CharlesChiuGit:
Other than that, I think it works as expected.

ayamir:
> 圖片
> 
> Also `lualine`'s `cwd` doesn't show on windows.

This line: https://github.com/ayamir/nvimdots/blob/c7caa6699c0ebb9f08601d32aaec4a261090b461/lua/modules/ui/config.lua#L491

Also because window has different file structure, use `home` as abbreviated prefix is not enough, will fix it.

Jint-lzxy:
> After some searching, I think this issue is unsolvable. Maybe just note it in the wiki. Non-sqlite-related `telescope` functions would work fine even with this error.

IMO this is because `sqlite3` is NOT in `package.cpath` _(Note the `c` here, not `package.path`)_. Will look into this later.

Jint-lzxy:
@CharlesChiuGit Can you set `vim.g.sqlite_clib_path = <your-path>` to your `sqlite3.dll`'s location to see if this helps? _(Note: Use `/` as path separator. DO NOT use `\\`)_

CharlesChiuGit:
> @CharlesChiuGit Can you set `vim.g.sqlite_clib_path = <your-path>` to your `sqlite3.dll`'s location to see if this helps? _(Note: Use `/` as path separator. DO NOT use `\\`)_

will test once i got home.

CharlesChiuGit:
I notice a small problem, some character in `ps1` script seems invalid when I install it now. Did sth changed?

Jint-lzxy:
@CharlesChiuGit That's weird, nothing changed.

What do you see:
```pwsh
1.
Write-Host '¯'

2.
[System.BitConverter]::ToUint32([System.Text.Encoding]::UTF32.GetBytes('¯'))
```

CharlesChiuGit:
> What do you see:
> 
> ```powershell
> 1.
> Write-Host '¯'
> 
> 2.
> [System.BitConverter]::ToUint32([System.Text.Encoding]::UTF32.GetBytes('¯'))
> ```

圖片

CharlesChiuGit:
Can confirm tt's not the font's issue, I check with windows terminal's default font: `Campbell`.

圖片

Jint-lzxy:
Hmm, that's weird _(Unicode literals are correctly interpreted on local machine)_. There might be some issues with github's `raw` data transfer. A recently merged PR #467 should refresh the cache, check if there is any issue now?

Jint-lzxy:
If that also failed then we might need to do it the "hard" way, using objects:
```pwsh
Set-ExecutionPolicy Bypass -Scope Process -Force; Invoke-Expression ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/ayamir/nvimdots/HEAD/install/install.ps1'))
```

Jint-lzxy:
If this also failed then I'll look into the weird powershell code tomorrow :(

CharlesChiuGit:
> Hmm, that's weird _(Unicode literals are correctly interpreted on local machine)_. There might be some issues with github's `raw` data transfer. A recently merged PR #467 should refresh the cache, check if there is any issue now?

With `Invoke-WebRequest 'https://raw.githubusercontent.com/ayamir/nvimdots/HEAD/install/install.ps1' | Invoke-Expression`:
圖片

With `Invoke-WebRequest 'https://raw.githubusercontent.com/CharlesChiuGit/nvimdots-ayamir/HEAD/install/install.ps1' | Invoke-Expression`:
圖片

Jint-lzxy:
Powershell probably failed to interpret the byte stream as UTF-8 literals :( Might also need some environment tweaks before calling the script

Jint-lzxy:
@ayamir We need to change the download link to:
```pwsh
Set-ExecutionPolicy Bypass -Scope Process -Force; Invoke-Expression ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/ayamir/nvimdots/HEAD/install/install.ps1'))
```

CharlesChiuGit:
> > 圖片
> > Also `lualine`'s `cwd` doesn't show on windows.
> 
> This line:
> 
> https://github.com/ayamir/nvimdots/blob/c7caa6699c0ebb9f08601d32aaec4a261090b461/lua/modules/ui/config.lua#L491
> 
> Also because window has different file structure, use `home` as abbreviated prefix is not enough. fixed at [09b2c74](https://github.com/ayamir/nvimdots/commit/09b2c745d3414fdf89f6b008889e3775efa079b2)

Can confirmed it's fixed.
圖片

Jint-lzxy:
@CharlesChiuGit https://github.com/ayamir/nvimdots/commit/f69d1a09abf5e2b4ed3592d91ab27f51b8d67e4e put all the instructions inside `_main` and then called that function to prevent unexpected behaviors. Hope it doesn't break anything :)

If you have time, can u help me test this commit? Thanks a lot  ❤️

CharlesChiuGit:
> @CharlesChiuGit Can you set `vim.g.sqlite_clib_path = <your-path>` to your `sqlite3.dll`'s location to see if this helps? _(Note: Use `/` as path separator. DO NOT use `\\`)_

圖片

I set the path with python's `sqlite3.dll` like so:
```lua
	-- fix sqlite3 missing-lib issue on Wondows
	if global.is_windows then
		vim.g.sqlite_clib_path = "C:/Users/Charles/scoop/apps/python/3.11.1/DLLs/sqlite3.dll"
	end
```
But once I called `:Telescope`, nvim force quite(so the new prompt show a red arrow.)

Did I use the wrong path? Or I should use other dll?

Jint-lzxy:
@CharlesChiuGit You should use the dll from the `sqlite` package, or just grab that from https://www.sqlite.org/download.html _("Precompiled Binaries for Windows")_.

CharlesChiuGit:
> @CharlesChiuGit [f69d1a0](https://github.com/ayamir/nvimdots/commit/f69d1a09abf5e2b4ed3592d91ab27f51b8d67e4e) put all the instructions inside `_main` and then called that function to prevent unexpected behaviors. Hope it doesn't break anything :)
> 
> If you have time, can u help me test this commit? Thanks a lot ❤️

First try:
The new commit doesn't break anything, everything works fine.

Second try, I intentionally open a nvim buffer when running the script, so the `nvim` folder will be lock down by OS file system.
The script breaks with following errors.
圖片

Did u mean "unexpected behaviors" like this?

CharlesChiuGit:
> @CharlesChiuGit You should use the dll from the `sqlite` package, or just grab that from [sqlite.org/download.html](https://www.sqlite.org/download.html) _("Precompiled Binaries for Windows")_.

I install `sqlite` via `scoop` and `scoop` doesn't contain any DLLs for sqlite.
What's download by scoop:
圖片

So I use fzf to find anything might relate to `sqlite3.dll` and find that there's a `sqlite3.dll` in the python's folder, which seems quite reasonable to use.
I didn't grab anything form [sqlite.org/download.html](https://www.sqlite.org/download.html) tho.

CharlesChiuGit:
Omg! I grab the `sqlite3.dll` from `sqlite.org` and it works!

圖片

But there's still some minor bugs when using telescope which might not be directly related to sqlite, i guess.

I think the force quite is probably related to mismatch version of `python's sqlite` and `scoop's sqlite`.

ayamir:
> @ayamir We need to change the download link to:
> 
> ```powershell
> Set-ExecutionPolicy Bypass -Scope Process -Force; Invoke-Expression ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/ayamir/nvimdots/HEAD/install/install.ps1'))
> ```

Updated in [zhihu](https://zhuanlan.zhihu.com/p/382092667) too.

CharlesChiuGit:
`:Telescope fzf` still having errors, even with `scoop install fzf`:
圖片

`:Telescope zoxide`, even with `scoop install zoxide`:
圖片

`:Telescope man_pages`, this is acceptable on Windows since it don't have `man`
圖片

```
$ fzf --version
0.37.0 (2023012)

$ zoxide --version
zoxide v0.9.0

**Comment by Jint-lzxy** [comment]
> Did u mean "unexpected behaviors" like this?

Yeah, in fact, this is the _correct_ behavior - sometimes the install script will exit without displaying the error message.

**Comment by Jint-lzxy** [comment]
> > @CharlesChiuGit You should use the dll from the `sqlite` package, or just grab that from [sqlite.org/download.html](https://www.sqlite.org/download.html) _("Precompiled Binaries for Windows")_.
> 
> I install `sqlite` via `scoop` and `scoop` doesn't contain any DLLs for sqlite. What's been downloaded by scoop: 圖片
> 
> So I use fzf to find anything might relate to `sqlite3.dll` and find that there's a `sqlite3.dll` in the python's folder, which seems quite reasonable to use. I didn't grab anything form [sqlite.org/download.html](https://www.sqlite.org/download.html) tho.

We need them both! If the package manager does not provide that `dll`, ppl can download it manually

**Comment by Jint-lzxy** [comment]
> `:Telescope fzf` still having errors, even with `scoop install fzf`: 圖片
> 

Nope `:Telescope fzf` is not a valid command. It replaces an **internal** function that handles the sorting. We should never use that directly.

> `:Telescope zoxide`, even with `scoop install zoxide`:

Same here, try `:Telescope zoxide list`

> `:Telescope man_pages`, this is acceptable on Windows since it don't have `man`

😄

**Comment by CharlesChiuGit** [comment]
> Nope `:Telescope fzf` is not a valid command. It replaces an internal function that handles the sorting. We should never use that directly.

Aha, got it. I tested it b/c `wilder` has a candidate for it. Personally I never use this command lol

> Same here, try `:Telescope zoxide list`

Indeed, that works.
圖片

No further issues on Windows then!

**Comment by CharlesChiuGit** [comment]
For `sqlite3.dll`, either we can add a note in installation script and wiki, or we can download the same version of DLL as the `sqlite/sqlite3` installed via pkg manager.

**Comment by CharlesChiuGit** [comment]
> Then our next step is to better classify the plugins 👍

Just about to say that lol

**Comment by Jint-lzxy** [comment]
> For `sqlite3.dll`, either we can add a note in installation script and wiki, or we can add into the script to download the same version of DLL as the version of `sqlite/sqlite3` installed via pkg manager.

I prefer to let users decided where to install that. ppl has different preferences lol

**Comment by Jint-lzxy** [comment]
> For `sqlite3.dll`, either we can add a note in installation script and wiki, or we can add into the script to download the same version of DLL as the version of `sqlite/sqlite3` installed via pkg manager.

Done: [Wiki: Issues#cannot-load-module-libsqlite3-or-similar-issues-on-windows](https://github.com/ayamir/nvimdots/wiki/Issues#cannot-load-module-libsqlite3-or-similar-issues-on-windows) and https://github.com/ayamir/nvimdots/pull/458/commits/538c34bea01c4ec67e031e0c6fd03307bd838278

**Comment by ayamir** [comment]
First of all, a clearer definition of current different categories is needed.
`completion` contains plugins about code completion. `ui` contains plugins which change the editor's interface. `editor` contains plugins which improve the default ability of vanilla `nvim`. `tools` contains plugins which use external tools to provide new ability for `nvim`.

**Comment by CharlesChiuGit** [comment]
圖片
圖片
圖片
圖片
圖片

plugin-structure

**Comment by CharlesChiuGit** [comment]
Will do some in-depth plugin dependence graph later. Currently I just write down from what the code says as it.

**Comment by ayamir** [comment]
Maybe we need to merge it as soon as possible if no classifications are proposed.

**Comment by Jint-lzxy** [comment]
> Maybe we need to merge it as soon as possible if no classifications are proposed.

Agreed. But we should first update wiki and release notes.

IMO there is no problem with the current structure _(Though it requires some explanation, it will be very clear after reading the documentation)_. We could consider adding `after/ftplugin` and `after/syntax`.

**Comment by CharlesChiuGit** [comment]
My propose:
- `completion` contains plugins about code completion.  
- `ui` contains plugins renders the interface without any actions after one fires `nvim`.  
- `editor` contains plugins which improve the default ability of vanilla `nvim`.  
- `tools` contains plugins using external tools and changing default layout to provide new ability for `nvim`.  
- `lang` contains plugins which relates to certain programming language.

Will provide a list of plugins should be moved shortly after.

**Comment by ayamir** [comment]
> But we should first update wiki and release notes.

We can add wiki pages as files in this repo like `lua/wiki/usage.md`. Maybe it's ok to add them in this pr.

**Comment by ayamir** [comment]
> Will provide a list of plugins those should be moved in my own sense shortly after.

Agree.

**Comment by ayamir** [comment]
> My propose:
> 
> * `completion` contains plugins about code completion.
> * `ui` contains plugins renders the interface **without any actions** after one fires `nvim`.
> * `editor` contains plugins which improve the default ability of vanilla `nvim`.
> * `tools` contains plugins using external tools and **changing default layout** to provide new ability for `nvim`.
> * `lang` contains plugins which relates to certain programming language.
> 
> Will provide a list of plugins those should be moved in my own sense shortly after.

Agree. We should write it in wiki.

**Comment by CharlesChiuGit** [comment]
Also, I think we should use `tool` instead of `tools`.

**Comment by CharlesChiuGit** [comment]
The plugin I think should change it's locations: 
```
- completion  
	- not thing to change  
- editor  
	- neoscroll -> ui  
	- smartyank -> tool  
	- specs -> ui  
	- toggleterm -> tool  
	- dap -> tool  
- lang  
	- not thing to change  
- tool  
	- not thing to do  
- ui  
	- nvim-tree -> tool  
```

If u guys thinks it's reasonable, I'll send a PR then.

CharlesChiuGit:
> We can add wiki pages as files in this repo like `lua/wiki/usage.md`. Maybe it's ok to add them in this pr.

Will do!

Jint-lzxy:
Above changes LGTM 👍

ayamir:
> - neoscroll -> ui  
> 	- smartyank -> tool  
> 	- specs -> ui  
> 	- toggleterm -> tool  
> 	- dap -> tool

ok.

CharlesChiuGit:
Just updated, plz check.

Will add wiki\usage.md later.

Jint-lzxy:
@CharlesChiuGit u also need to modify the four `require`s in `dap/init.lua`.

CharlesChiuGit:
> @CharlesChiuGit u also need to modify the four `require`s in `dap/init.lua`.

I only found three?
圖片

CharlesChiuGit:
圖片

Side note: I seem to have an error related to telescope's picker, any idea?

Jint-lzxy:
> I only found three?

Oh god I missed a commit. Sorry for that

Jint-lzxy:
> 圖片
> 
> Side note: I seem to have an error related to telescope's picker, any idea?

We need the whole stack traceback, try `:messages`

CharlesChiuGit:
> We need the whole stack traceback, try `:messages`

Aha, I found the issue, for some reason `ripgrep` get uninstalled. Not sure why. But after scoop install it again, telescope works fine.

Jint-lzxy:
> Aha, I found the issue, for some reason `ripgrep` get uninstalled, and the win-scripts show it's installed, not sure why? But after scoop install it again, telescope works fine.

Hmm maybe it's just telescope can't find `rg` in `$PATH` / keyboard input during startup will interrupt the current loading operation, this happens occasionally.
***
@CharlesChiuGit Also a side note here. You can add your email address _(`Charles@Dino-Lite.com` from commit info)_ [on this page](https://github.com/settings/emails) to better attribute your commit _(at present, we can't get who the commit having a default avatar belongs to at a glance 😃)_.

CharlesChiuGit:
> Also a side note here. You can add your email address _(`Charles@Dino-Lite.com` and `charleschiu2012@gmail.com` from commit info)_ [on this page](https://github.com/settings/emails) to better attribute your commit _(at present, we can't get who the commit having a default avatar belongs to at a glance 😃)_.

I think I did add my emails to github long time ago, but for some reason github always shows sth like `
Charles Chiu authored and Charles Chiu committed XXX mins aga.` 

圖片
圖片

CharlesChiuGit:
I'm holding my `wiki\plugins.md` commit for now if this problem can be addressed. 🤣

CharlesChiuGit:
Also, I think we should change `install/` to `scripts` and `my-sniipets` to `snips`. 
How do u guys think?

Jint-lzxy:
> I'm holding my `wiki\plugins.md` commit for now if this problem can be addressed. 🤣

Try directly pasting the output of `git config --global user.email`?

Jint-lzxy:
> > I'm holding my `wiki\plugins.md` commit for now if this problem can be addressed. 🤣
> 

Try directly pasting the output of `git config --global user.email` to GitHub?

CharlesChiuGit:
> Try directly pasting the output of `git config --global user.email` to GitHub?

Omg, git doesn't pickup my `.gitconfig`!?
圖片

Jint-lzxy:
Probably the location of `.gitconfig` is incorrect :(

Jint-lzxy:
Commit information is parsed correctly now 🥳

ayamir:
> Also, I think we should change `install/` to `scripts/` and `my-snipets/` to `snips/`. How do u guys think?

ok

CharlesChiuGit:
Omg it works. TBO, the behavior is super weird. 
1. I delete my `C:\Users\charles\.gitconfig`, which should be the place where `.gitconfig` on windows stores, normally.
2. I reconfigure `git config --global` in command line.
3. `git-for-windows` creates an `Z:\` disk section on my work machine and store the `.gitconfig` there??? That's sth I never seen before, so weird.

But hey, at least it's working lol

CharlesChiuGit:
I think this PR is ready to merged. @Jint-lzxy @ayamir PTAL.

Jint-lzxy:
I will merge this PR after modifying `README` and publishing a release.

CharlesChiuGit:
I'll update the https://github.com/ayamir/nvimdots/wiki/Plugins then.

ayamir:
`my-snippets` in `snips/package.json` should be renamed to `snips`.

CharlesChiuGit:
Personally, I think https://github.com/ayamir/nvimdots/wiki/Plugins shouldn't use table as a display format, and there's not need to explain the effect of each plugin. ppl can just click on the links and read it's README.

I think using indents to display the relations of the plugins would be better, like in https://github.com/ayamir/nvimdots/pull/458#issuecomment-1414880079.

ayamir:
> Personally, I think https://github.com/ayamir/nvimdots/wiki/Plugins shouldn't use table as a display format, and there's not need to explain the effect of each plugin. ppl can just click on the links and read it's README.

Whether to remove the explanation of the effect or not is optional. The purpose I add it is explain the plugins' function briefly.

> I think using indents to display the relations of the plugins would be better, like in [#458 (comment)](https://github.com/ayamir/nvimdots/pull/458#issuecomment-1414880079).

Maybe this format is clearer? It's ok.

ayamir:
> We could consider adding `after/ftplugin` and `after/syntax`.

This change be introduced in #459.

Jint-lzxy:
@CharlesChiuGit I plan to put `plugins.md` directly to wiki. This way users can get all information in one place.

> I think using indents to display the relations of the plugins would be better, like in https://github.com/ayamir/nvimdots/pull/458#issuecomment-1414880079.

We can adopt those changes using `- **<plugin-name> (bold)**: <simple-description>`

ayamir:
> We can adopt those changes using - **<plugin-name> (bold)**: <simple-description>

So maybe `CONTRIBUTING.md` and commit style check should be created?

CharlesChiuGit:
> So maybe `CONTRIBUTING.md` and commit style check should be created?

Could be helpful if there's new plugins coming from PRs.

ayamir:
I will implement it in next pr if ok.

Jint-lzxy:
https://github.com/ayamir/nvimdots/wiki/Usage. I rewrote the first two parts. Hope this can be easily understood @CharlesChiuGit @ayamir

ayamir:
> https://github.com/ayamir/nvimdots/wiki/Usage. I rewrote the first two parts. Hope this can be easily understood @CharlesChiuGit @ayamir

We can set `user` to `custom` if we use `custom` in the code.

Jint-lzxy:
> We can set `user` to `custom` if we use `custom` in the code.

Updated.

Jint-lzxy:
@ayamir External links also need to be updated. See https://github.com/ayamir/nvimdots/pull/458/commits/36f331f4303f2102644050b5f60e1e5a85349078

ayamir:
Done

Jint-lzxy:
@CharlesChiuGit IMO it is unnecessary to introduce two empty files. Users can customize them according to the wiki.

ayamir (approved):
Can be merged once the new release note done.

PR #459, opened 2023-02-01
feat: add rust crates.io support

ClSlaid:
this should be the missing part of rust support

it's useful for checking dependencies in rust projects

Signed-off-by: 蔡略 <cailue@bupt.edu.cn>


CharlesChiuGit:
Hi, we're now doing some major refactoring in #458 . We might need u to adjust quite few things after that PR is merged.

lua/modules/lang/config.lua
```diff
@@ -185,6 +185,151 @@ function config.rust_tools()
...
+				feature = "  %s",
+				enabled = " %s",
+				transitive = " %s",
+				normal_dependencies_title = " Dependencies",
+				build_dependencies_title = " Build dependencies",
+				dev_dependencies_title = " Dev dependencies",
+				dependency = "  %s",
+				optional = " %s",
+				dependency_version = "  %s",
+				loading = "  ",
```
Jint-lzxy, line 264: Looks great!  But before merging, can you replace the icons in this commit with those from the library?

Documentation: [Wiki: Usage#icons](https://github.com/ayamir/nvimdots/wiki/Usage#icons). It's recommended that you pass `add_space` as `true` to automatically add a trailing space. 

Example _(line 4, 10 and 11)_:
https://github.com/ayamir/nvimdots/blob/2b1cff195a3ce8c46d5d3294959b707de747656c/lua/modules/tools/config.lua#L3-L11

lua/modules/lang/config.lua
```diff
@@ -185,6 +185,151 @@ function config.rust_tools()
...
+				goto_item = { "gd", "K", "<C-LeftMouse>" },
+				jump_forward = { "<c-i>" },
+				jump_back = { "<c-o>", "<C-RightMouse>" },
+			},
+		},
+		src = {
+			insert_closing_quote = true,
+			text = {
+				prerelease = "  pre-release ",
+				yanked = "  yanked ",
```
Jint-lzxy, line 318: Also these two lines

Jint-lzxy (changes requested):

ClSlaid:
@Jint-lzxy all done, PTAL.

lua/modules/lang/config.lua
```diff
@@ -185,6 +185,111 @@ function config.rust_tools()
...
+		autoupdate = true,
+		autoupdate_throttle = 250,
+		loading_indicator = true,
+		date_format = "%Y-%m-%d",
+		thousands_separator = ".",
+		notification_title = "Crates",
+		curl_args = { "-sL", "--retry", "1" },
+		disable_invalid_feature_diagnostic = false,
+		text = {
+			loading = " " .. icons.diagnostics.Perf .. "Loading",
```
Jint-lzxy, line 210: @ClSlaid https://github.com/ayamir/nvimdots/pull/459/commits/e6df04d761d4d59ae78e1c778d09252a5920b13b fixed function calls but a simple inspection showed that some icons are not defined under current classification like this one. Please ensure that neovim can run without errors before pushing changes

Jint-lzxy (changes requested):

Jint-lzxy:
@ClSlaid Sorry to inform you, but all PRs related to the new plugins/features would be merged after #458 at the very least. As mentioned by CharlesChiuGit, you may need to make some additional changes. Apologize for the inconvenience.

> cc @Jint-lzxy need suggestions on key maps, it this key mapping OK? PTAL.

I'm afraid I can't give you an answer now. We will evaluate the possible changes/breakages after #458. But my opinion now is that it is not appropriate to add a huge number of **global** keybindings **only applicable to a specific language** for the base configuration. Perhaps it would be a better choice to add local keybindings when specific conditions are met. _(AFAIK many people will use `<leader>c*` to do some other stuffs)_

Also a quick note here, `"BufRead Cargo.toml"` is not a valid event name. See `:h events` for more information.

ClSlaid:
@Jint-lzxy refactored key-mappings to buffer specific. PTAL.

Jint-lzxy:
@ClSlaid I just added a `ftplugin` that could load those keymaps according to the filetype, and also modified some of them so that they do not conflict with the base config.

I don't use `rust`, so there are a couple of issues that I cannot debug/confirm before merging:
1. When would you usually require this plugin? Is this useful for all `rust` files or just `Cargo.toml`?
2. I used regex to migrate those keymaps, so I would be most grateful if you could test them again 👍
```regex
:'<,'>s/\tvim.keymap.set("\(.\)",\s*"\(.*\)".\s*function()\n\t*\(.*\)$\n\t*.*})/\1noremap <silent> <buffer> \2 :lua \3<CR>
```
4. Why `show_*_popup()` is followed by `show_popup()`? Isn't `show_*_popup()` enough to show a popup window?

CharlesChiuGit:
@Jint-lzxy I think it's fine to use `lua` in `after/plugin`. Here's some [examples](https://github.com/CharlesChiuGit/nvimdots.lua/tree/main/after/ftplugin). Or maybe @ClSlaid can modify `rust.vim` to into `rust.lua` using `vim.api.nvim_set_keymap`.

Jint-lzxy:
@CharlesChiuGit Just wanna make this file a little shorter 😄 _(19 lines v.s. 67 lines)_. It is also more intuitive to use vimscript when it comes to modifying options.

CharlesChiuGit:
Understandable, got it!

ClSlaid:
> When would you usually use this plugin? Is this useful for all rust files or just Cargo.toml?

The `Cargo.toml` is a manifest file containing metadata of your rust projects, and this plugin could show the metadata of your project dependencies. So this file is only needed for `Cargo.toml`.

> Why show_*_popup() is followed by show_popup()? Isn't show_*_popup() enough to show a popup window?

To make sure that the pop-up window will surely been shown.

> Just wanna make this file a little shorter 😄 (19 lines v.s. 67 lines). It is also more intuitive to use vimscript when it comes to modifying options.

I prefer keeping it in Lua, since the project claims itself to be a pure Lua configuration.

Jint-lzxy:
Got it! Thanks for clarifying that ❤️

Only one last thing - can you complete the description for [each keymap](https://github.com/ayamir/nvimdots/pull/459/files#diff-58283cf08d3a239ac1f4134d454e63cf04e227f67feec706a0123acde5f333a6R11)? The format is as follows:
```lua
:with_desc("crates: <a-brief-desc>"),
```

Jint-lzxy:
@ClSlaid Got it! Thanks for clarifying that ❤️

Only one last thing - can you complete the description for [each keymap](https://github.com/ayamir/nvimdots/pull/459/files#diff-58283cf08d3a239ac1f4134d454e63cf04e227f67feec706a0123acde5f333a6R11)? The format is as follows:
```lua
:with_desc("crates: <a-brief-desc>"),
```

ClSlaid:
@Jint-lzxy should be done, PTAL.

PR #465, opened 2023-02-02
feat(editor): add whitespace plugin for trim file

Groveer:


Jint-lzxy:
IMO there is no need to add this plugin. All code formatters _should_ support this feature, and I would like the base configuration to be kept as simple as possible, especially when it comes to adding plugins that cannot be lazily loaded.

But anyway, thank you for the contribution 👍

PR #467, opened 2023-02-02
fix(script-win): Typo

Jint-lzxy:
Also a side note:
> 使用chocolatey安装必要的依赖，下载相应的安装包时会检测系统代理，因为大多是从github下载，所以最好开启系统代理。
> _(From `zhihu`)_

In fact, users can choose to use `chocolatey` or `scoop` to fetch deps. TBH, I see more people using `scoop` so I decided to support both of them.

IMO it would be better to address that the script also supports `scoop` _(and users must install one of them before invoking the script)_, otherwise it may lead to confusion. _(ppl probably won't read the code before installation)_

### DIFF
--- a/install/install.ps1
+++ b/install/install.ps1
@@ -129,7 +129,7 @@ function Check-Def-Exec ([Parameter(Mandatory = $True)][ValidateNotNullOrEmpty()
 
 function Query-Pack {
 	if ((Check-Def-Exec -WithName "scoop") -and (Check-Def-Exec -WithName "choco")) {
-		prompt -Msg "   [Detected] Multiple package mgrs detected."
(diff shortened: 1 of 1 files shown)


ayamir:
> Also a side note:
> 
> > 使用chocolatey安装必要的依赖，下载相应的安装包时会检测系统代理，因为大多是从github下载，所以最好开启系统代理。
> > _(From `zhihu`)_
> 
> In fact, users can choose to use `chocolatey` or `scoop` to fetch deps. TBH, I see more people using `scoop` so I decided to support both of them.
> 
> IMO it would be better to address that the script also supports `scoop` _(and users must install one of them before invoking the script)_, otherwise it may lead to confusion. _(ppl probably won't read the code before installation)_

ok

PR #475, opened 2023-02-06
fix(script-win): Avoid prompting for user input when failed

Jint-lzxy:
This should eliminate that weird message showing:
```console
cmdlet prompt at command pipeline position 1

Supply values for the following parameters:
Msg:
```

@ayamir @CharlesChiuGit  I would be most grateful if you could test this fix _(let the program exit abnormally, calling `_abort`)_

### DIFF
--- a/install/install.ps1
+++ b/install/install.ps1
@@ -5,8 +5,6 @@
 
 Set-StrictMode -Version 3.0
 
-$ErrorActionPreference = "Stop" # Exit when command fails
-
 # global-scope vars
 $REQUIRED_NVIM_VERSION = [version]'0.8.0'
 $USE_SSH = $True


ayamir:
No help currently:
image

CharlesChiuGit:
I think @Jint-lzxy need us to edit the script a bit like:
```ps1
...

	_abort -Msg "test"
	Ring-Bell
	Wait-For-User
```

Jint-lzxy:
@ayamir It is intentional to issue such prompt when directly sending `SIGABRT` to terminate the script. _(This forces the user to close this window to clean up orphan processes, as we do not want them to continue executing the install script in this session)_.

One approach is to press keys other than `RETURN` when prompting for confirmation:
https://github.com/ayamir/nvimdots/blob/24f89efa6f8a19e3db5fba9c3cf2d09c1f2b9621/install/install.ps1#L73-L81

CharlesChiuGit:
I edited the file like this:
```ps1
...
	if ((Test-Path $env:CCDEST_DIR)) {
		warn -Msg "The destination folder: `"$env:CCDEST_DIR`" already exists."
		warn-Ext -Msg "We will make a backup for you at `"$env:CCBACKUP_DIR`"."
	}

	_abort -Msg "test" -Type "NotInstalled" -----> add this line
	Ring-Bell
	Wait-For-User
...
```
And the result:
圖片

If I understand correctly, I think this PR is working as expected.

ayamir:
> @ayamir It is intentional to issue such prompt when directly sending `SIGABRT` to terminate the script. _(This forces the user to close this window to clean up orphan processes, as we do not want them to continue executing the install script in this session)_.
> 
> One approach is to press keys other than `RETURN` when prompting for confirmation:
> 
> https://github.com/ayamir/nvimdots/blob/24f89efa6f8a19e3db5fba9c3cf2d09c1f2b9621/install/install.ps1#L73-L81

If so, it's ok.

ayamir (approved):

CharlesChiuGit:
@Jint-lzxy Actually, `ripgrep` is not being check by the script lol.

圖片

ayamir:
One question, can you install these packages successfully without administrator privilege?

CharlesChiuGit:
> One question, can you install these packages successfully without administrator privilege? I need to remind windows users in readme if so.

If u use `scoop`, then admin-previlege is not needed.
I'm not sure about choco tho, but to my understanding users need extra settings to install non-admin-previledge choco.

ayamir:
`choco` need it.

PR #476, opened 2023-02-06
chore(clangd): add auto-path completion for query-driver

aarnphm:
I added this small change for clangd to auto-complete the tools that is used for --query-driver

Maybe after custom config for users mature a bit more, we can allow them to configure such features as well, but for now if users need to add any new driver they can just add it in the table instead of modifying the specific path.

### DIFF
--- a/lua/modules/configs/completion/servers/clangd.lua
+++ b/lua/modules/configs/completion/servers/clangd.lua
@@ -23,7 +23,7 @@ local function switch_source_header_splitcmd(bufnr, splitcmd)
 	end
 end
 
-local get_binary_path = function(binary)
+local function get_binary_path(binary)
 	local path = nil
 	if global.is_mac or global.is_linux then
 		path = vim.fn.trim(vim.fn.system("which " .. binary))
@@ -36,7 +36,7 @@ local get_binary_path = function(binary)
 	return path
 end
 
-local get_binary_path_list = function(binaries)
+local function get_binary_path_list(binaries)
 	local path_list = {}
 	for _, binary in ipairs(binaries) do
 		local path = get_binary_path(binary)
@@ -46,6 +46,7 @@ local get_binary_path_list = function(binaries)
 	end
 	return table.concat(path_list, ",")
 end
+
 -- https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/clangd.lua
 return {
 	single_file_support = true,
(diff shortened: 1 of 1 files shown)


ayamir:
So the `clang++` which is not in the `PATH` can't be found in this way. Users should add where the `clang++` located to `PATH`.

aarnphm:
Hmm. Maybe we can also inform users to add such driver into PATH in the Wiki? What do you think?

lua/modules/configs/completion/servers/clangd.lua
```diff
@@ -22,6 +23,30 @@ local function switch_source_header_splitcmd(bufnr, splitcmd)
 	end
 end
 
+local function get_binary_path(binary)
+	local path = nil
+	if global.is_mac or global.is_linux then
+		path = vim.fn.trim(vim.fn.system("which " .. binary))
+	elseif global.is_windows then
+		path = vim.fn.trim(vim.fn.system("where " .. binary))
```
aarnphm, line 31: Also from the Windows folks my question is whether this is the correct way to find binary?
I don't have a Windows machine so I'm not entirely sure

ayamir: Test passwd for window.

ayamir:
> Hmm. Maybe we can also inform users to add such driver into PATH in the Wiki? What do you think?

OK. This commit also consider the problem caused by different OS. It's also intuitive that `clang++` should in the `PATH` for most users. but we also need consider users who just use multiple versions of `clang++` and note for them.

aarnphm:
I can update the Wiki if needed.

ayamir (approved):

CharlesChiuGit:
> I can update the Wiki if needed.

Tks! U can type what needs to added, and i'll add it later.

aarnphm:
ok. So I think we just need to inform users who use clangd to add any of the required `query-driver` to PATH on all platform.

CharlesChiuGit:
Do u have a draft that I can work on?

aarnphm:
I don't have one right now. I'm wondering if the repo should include a FAQ section 🤔

CharlesChiuGit:
https://github.com/ayamir/nvimdots/wiki/Issues

aarnphm:
cc @CharlesChiuGit https://github.com/ayamir/nvimdots/wiki/Issues#issues-while-using-clangd-cc-lsp

CharlesChiuGit:
Nice! So u have permissions to edit wikl?

ayamir:
> Nice! So u have permissions to edit wikl?

Everyone has the edit permission for convenience.

aarnphm:
Oh, is that for contributor only? or anyone from the internet can edit?

if it is the latter then ... 😄

PR #482, opened 2023-02-06
refactor: use on_attach to set mapping

aarnphm:
This PR refactor current gitsigns config, that for setting up key binding we can use `on_attach`

This makes better customization flow as we can access to vim.fn


ayamir:
A little issue is `which-key` won't show the command content after `<leader>h` being pressed in this way.

aarnphm:
ah I see. I updated a section in `keybinds` that allows setting up description
`:with_desc`

ayamir (approved):

PR #483, opened 2023-02-06
fix(install-win): Update exec name for 'ripgrep'

Jint-lzxy:
@CharlesChiuGit I'm not sure if this is intentional or not, but the name of the executable for `ripgrep` should be `rg` on all platforms.
```rust
app.gen_completions("rg", Shell::PowerShell, &outdir);
```
[BurntSushi/ripgrep/build.rs#L43](https://github.com/BurntSushi/ripgrep/blob/master/build.rs#L43)

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -10,8 +10,8 @@ $REQUIRED_NVIM_VERSION = [version]'0.8.0'
 $USE_SSH = $True
 
 # package mgr vars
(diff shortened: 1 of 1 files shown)


ayamir (approved):
Indeed, I realized this problem after approved lol.

CharlesChiuGit:
Aha, sorry about that. tks!

PR #485, opened 2023-02-06
chore(keymap): use telescope keymaps to provide command panel.

ayamir:
I find that `which-key` won't work without `with_desc()` setting for keymaps. So I add descriptions for almost of all of keymaps. What's more, I find that `:Telescope keymaps` implements almost the same feature of `legendary.nvim` and former can also utilize args passed to `with_desc()`. So I decide to remove `legengary.nvim` and `dressing.nvim` to enhance cohesion and reduce coupling.


ayamir:
I use `i`, `n`, `v` and `c` as the suffix of `edit` to mark the execute mode of the keymap.

Jint-lzxy (approved):
🍰

CharlesChiuGit (commented):
LGTM

PR #488, opened 2023-02-07
feat(telescope): add grep_string keymap

Groveer:


lua/keymap/init.lua
```diff
@@ -231,6 +231,7 @@ local plug_map = {
 		:with_silent()
 		:with_desc("editn: Change current direrctory by zoxide"),
 	["n|<leader>fb"] = map_cu("Telescope buffers"):with_noremap():with_silent():with_desc("find: Buffer opened"),
+	["n|<leader>fs"] = map_cu("Telescope grep_string"):with_noremap():with_silent():with_desc("find: Current word"),
```
aarnphm, line 234: qq: would live_grep_args achive the same?

Groveer: no, live_grep_args must input args, this command-line will grep the string of the current cursor

Groveer: It will help me to find strings quickly

Jint-lzxy: @aarnphm `grep_string` is equivalent to passing the string under cursor as the first parameter to `live_grep_args`. User input inside that prompt is equivalent to appending additional parameters to `live_grep_args`.

Jint-lzxy (approved):
LGTM 🎐

lua/keymap/init.lua
aarnphm: ah got it.

PR #495, opened 2023-02-09
refactor(keymap): refactor plugin keymaps according to scopes

CharlesChiuGit:


aarnphm:
food for thought: I'm thinking about moving per plugin mapping to its corresponding config. So that the default keymap should only concern all of the basic movement and functionality from nvim that is independent of the plugin system.

CharlesChiuGit:
> food for thought: I'm thinking about moving per plugin mapping to its corresponding config init. So that the default keymap should only concern all of the basic movement and functionality from nvim that is independent of the plugin system.
> 
> ```lua
> editor["plugin"] = {
>   lazy = true
>   init = function()
>      local k = require "keymap.bind"
>      k.nvim_load_mapping {
>      ["..."] = ...
>      }
>   end
> }
> ```
> 
> It probably provide a better separation of concern.

I personally love this idea! I once done this with my own config. 
Two major issue might need to be addressed:
1. It will be a bit hard to debug. Since we might not use this config as our live nvim config, so we can't call `<C-p>` to see all existing keymaps.
2. This would make `plugins\{scope}.lua` super long, we should probably avoid that. We might need a `lua/modules/keymaps/{scopes}/{plugin-name}.lua` to avoid this.

Maybe we can come out with a better idea to solve above issues.

aarnphm:
> It will be a bit hard to debug or find duplicated keymaps. Since we might not use this config as our live nvim config, so we can't call <C-p> to see all existing keymaps.

Which-keys should solve this for us during `checkhealth` right?

> This would make plugins\{scope}.lua super long, we should probably avoid that. We might need a lua/modules/keymaps/{scopes}/{plugin-name}.lua to avoid this.
Agree with this point.

I don't like what I'm about to say, but `lua/modules/config/{scope}/{plugin}.lua` should takes a table, and then when users specify under `lua/modules/plugins/{scope}/{plugin}.lua` we can have something such as:

```lua
editor['plugin'] = require("modules.utils").create_config({lazy = true, ...}, require("editor.plugin"))
```

the default behaviour is all field under `require('editor.plugin')` will overwrite existing mapping provided to `create_config`

aarnphm:
I just realize that `create_config` is just vim.tbl_extend :)

Jint-lzxy:
@aarnphm I've once considered this idea, but IMO it would be better for users to query all keymaps in one place _(especially for those who are not familiar with the project structure - they may define keymaps in the wrong place if we put those keymaps into different files)_, and reduce merge conflicts as much as possible _(the options for each spec should be left to users, and as much as possible)_

It is indeed a good idea, but we should consider novices who are new to `neovim`. The ability to run out-of-box and easy to make modifications is also consequential.

Just some thoughts.

CharlesChiuGit:
> Which-keys should solve this for us during `checkhealth` right?

I'm not sure about that, I never got `which-key`'s `checkhealth` worked as expected.

aarnphm:
@Jint-lzxy makes sense. I think that the current refactor should achieve this as well.

Jint-lzxy:
> I don't like what I'm about to say, but `lua/modules/config/{scope}/{plugin}.lua` should takes a table, and then when users specify under `lua/modules/plugins/{scope}/{plugin}.lua` we can have something such as:
> 
> ```lua
> editor['plugin'] = require("modules.utils").create_config({lazy = true, ...}, require("editor.plugin"))
> ```
> 
> the default behaviour is all field under `require('editor.plugin')` will overwrite existing mapping provided to `create_config`
> 
> ```lua
> M.create_config = function(opts, override)
> ```
> 
> My only concern with `lua/modules/keymaps/{scopes}/{plugin-name}.lua` is that when you do file grep it will have like 3-4 files with the same name 😆

@aarnphm `lazy.nvim` provides [this feature](https://github.com/folke/lazy.nvim#-structuring-your-plugins), but FWIW it would be more difficult for new users to get started. See: [Wiki: Usage#add-a-new-plugin](https://github.com/ayamir/nvimdots/wiki/Usage#add-a-new-plugin). Telling users to add specs in one place and their own configuration in another place is much simpler IMO. Otherwise, we also need to explain how spec merging works, and it is likely that users will make mistakes when working with nested tables - TBH this could lead to countless issues 🤦‍♂️

CharlesChiuGit:
> TBH this could lead to countless issues 🤦‍♂️

Lmao, this one does put a smile on my face.

Jint-lzxy:
> @Jint-lzxy makes sense. I think that the current refactor should achieve this as well.

@aarnphm Feel free to open an issue to discuss a _more_ appropriate "compromise" 😄 It's hard to "meet some sort of balance" 🤣

CharlesChiuGit:
https://github.com/ayamir/nvimdots/discussions is another gd place to start this kind of discussions. TBH, it's a bit under used.

CharlesChiuGit:
Just found a stupid mistake. https://github.com/ayamir/nvimdots/commit/4d38042b9993b7792c02801223a84dc067740533 should be  
> rename: rename `pbind` to `bind` since it's not only for plugins

Jint-lzxy:
Fixed :)

CharlesChiuGit:
@Jint-lzxy omg, I don't ppl can do that! Any resources to learn this kind of magic?

aarnphm:
> @aarnphm Feel free to open an issue to discuss a _more_ appropriate "compromise" 😄 It's hard to "meet some sort of balance" 🤣

that's the million dollar question in software ain't it 😄

Jint-lzxy:
@CharlesChiuGit It's all about interactive rebasing and force-push :) See:

https://stackoverflow.com/questions/5509543/how-do-i-properly-force-a-git-push
https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

CharlesChiuGit:
@aarnphm Actually I very curious about how u structure your nvimdots. Mind sharing the link?

CharlesChiuGit:
> @CharlesChiuGit It's all about interactive rebasing and force-push :) See:
> 
> [stackoverflow.com/questions/5509543/how-do-i-properly-force-a-git-push](https://stackoverflow.com/questions/5509543/how-do-i-properly-force-a-git-push) [git-scm.com/book/en/v2/Git-Tools-Rewriting-History](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History)

Appreciate that!

aarnphm:
> @aarnphm Actually I very curious about how u structure your nvimdots. Mind sharing the link?

https://github.com/aarnphm/editor my own fork of nvimdots 😄

CharlesChiuGit:
@Jint-lzxy Do u know what's 
```lua
	["o|m"] = map_callback(function()
		require("tsht").nodes()
	end):with_silent(),
```
It doesn't seem belong to any of current plugin, is this some legacy code?

ayamir:
> 

No, https://github.com/mfussenegger/nvim-treehopper

CharlesChiuGit:
> No, [mfussenegger/nvim-treehopper](https://github.com/mfussenegger/nvim-treehopper)

But this plugin is not listed in `plugins/editor.lua`. Do we need to add it back?

CharlesChiuGit:
@Jint-lzxy @aarnphm PTAL.
ayamir, I think it would be better not to bother u too much considering your current condition.

Jint-lzxy:
@CharlesChiuGit im a bit tied up at the moment, will push some changes later 👍

lua/modules/plugins/tool.lua
```diff
@@ -62,6 +62,7 @@ tool["gelguy/wilder.nvim"] = {
 tool["nvim-telescope/telescope.nvim"] = {
 	lazy = true,
 	cmd = "Telescope",
+	event = "UIEnter",
```
aarnphm, line 65:
```suggestion
	event = "LazyVimStarted",
```

I think Telescope can be loaded after UIEnter.

aarnphm (commented):
quick comment from me. Probably still need another look from Jint

lua/modules/plugins/tool.lua
aarnphm: oh nvm discard this comment

lua/keymap/tool.lua
```diff
@@ -0,0 +1,216 @@
...
+		:with_desc("lsp: Show workspace diagnostics"),
+	["n|<leader>tq"] = map_cr("TroubleToggle quickfix")
+		:with_noremap()
+		:with_silent()
+		:with_desc("lsp: Show quickfix list"),
+	["n|<leader>tl"] = map_cr("TroubleToggle loclist"):with_noremap():with_silent():with_desc("lsp: Show loclist"),
+
+	-- Plugin: telescope
+	["n|<C-p>"] = map_callback(function()
+			_Command_panel()
```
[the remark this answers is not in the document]
CharlesChiuGit: ys it does

lua/modules/configs/tool/toggleterm.lua
```diff
@@ -34,4 +34,40 @@ return function()
...
+		if vim.fn.executable("lazygit") then
+			local lazygit = Terminal:new({ cmd = "lazygit", hidden = true, direction = "float" })
+			lazygit:toggle()
+		else
+			not_found_notify("Lazygit")
+		end
+	end
+
+	function _Htop_toggle()
+		if vim.fn.executable("lazygit") then
```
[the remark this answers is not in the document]
CharlesChiuGit: aha, forgot to change that lol

lua/modules/plugins/tool.lua
aarnphm, line 65: I don't think it should be `UIEnter` here. It will slow down StartupTime.

Telescope should be invoked when `Telescope` cmd is used.

lua/modules/configs/tool/toggleterm.lua
```diff
@@ -34,4 +34,40 @@ return function()
...
 		shell = vim.o.shell, -- change the default shell
 	})
+
+	local function not_found_notify(program)
+		vim.notify(string.format("[%s] not found!", program), vim.log.levels.ERROR, { title = "toggleterm.nvim" })
+	end
+
+	local Terminal = require("toggleterm.terminal").Terminal
+
+	function _Lazygit_toggle()
```
aarnphm, line 44: Another note is that this remove cache object. So if user accidentally close the terminal (not exit), then the terminal won't be clean correctly.

Everytime this function is called, it will draw a new terminal windows instead.

aarnphm: Maybe a solution for this is to have a map for cache of all available cmd that can be toggled by ToggleTerm as a first step.

```lua
local cache = {lazygit=nil, htop=nil, python=nil}

...

if vim.fn.executable("lazygit") then
   if not cache.lazygit then 
      cache.lazygit = ...
   end
   cache.lazygit:toggle()
else
   ...
end
```

CharlesChiuGit: That's gd idea! Will do.

ayamir:
@aarnphm A collaborators invitation has sended for you, thanks for your contributions.

CharlesChiuGit:
I'm not sure if https://github.com/ayamir/nvimdots/commit/37f4f5e247c83f75c65f0cbeeaf0f98d9ad8f133 did cache toggleterm windoes or not. Not sure how to reproduce "accidentally close the terminal".

aarnphm:
I run into this issue when I forgot to close the terminal correct, so when trying to do `:q`, there is an error about a buffer still opening. 

It turns out to be the unclosed toggleterm buffer.

lua/modules/configs/tool/toggleterm.lua
```diff
@@ -34,4 +34,58 @@ return function()
...
+		python = nil,
+	}
+
+	function _Lazygit_toggle()
+		if toggleterm_cache.lazygit then
+			toggleterm_cache.lazygit:toggle()
+		else
+			if vim.fn.executable("lazygit") then
+				local lazygit = Terminal:new({ cmd = "lazygit", hidden = true, direction = "float" })
+				lazygit:toggle()
```
aarnphm, line 56:
```suggestion
				toggleterm_cache.lazygit = Terminal:new({ cmd = "lazygit", hidden = true, direction = "float" })
				toggleterm_cache.lazygit:toggle()
```

CharlesChiuGit:
> I run into this issue when I forgot to close the terminal correct, so when trying to do `:q`, there is an error about a buffer still opening.
> 
> It turns out to be the unclosed toggleterm buffer.

Actually, non of above new toggleterm windows(lazygit, htop, python) can use `:q` to create a accidentally closed terminal.

lua/modules/configs/tool/toggleterm.lua
```diff
@@ -34,4 +34,58 @@ return function()
...
+		end
+	end
+
+	function _Htop_toggle()
+		if toggleterm_cache.htop then
+			toggleterm_cache.htop:toggle()
+		else
+			if vim.fn.executable("htop") then
+				local htop = Terminal:new({ cmd = "htop", hidden = true, direction = "float" })
+				htop:toggle()
```
aarnphm, line 69:
```suggestion
				toggleterm_cache.htop = Terminal:new({ cmd = "htop", hidden = true, direction = "float" })
				toggleterm_cache.htop:toggle()
```

lua/modules/configs/tool/toggleterm.lua
```diff
@@ -34,4 +34,58 @@ return function()
...
+	function _Python_toggle()
+		if toggleterm_cache.python then
+			toggleterm_cache.python:toggle()
+		else
+			if vim.fn.executable("python3") then
+				local python = Terminal:new({ cmd = "python3", hidden = true, direction = "float" })
+				python:toggle()
+			elseif vim.fn.executable("python") then
+				local python = Terminal:new({ cmd = "python", hidden = true, direction = "float" })
+				python:toggle()
```
aarnphm, line 85:
```suggestion
				toggleterm_cache.python = Terminal:new({ cmd = "python3", hidden = true, direction = "float" })
				toggleterm_cache.python:toggle()
			elseif vim.fn.executable("python") then
				toggleterm_cache.python = Terminal:new({ cmd = "python", hidden = true, direction = "float" })
				toggleterm_cache.python:toggle()
```

aarnphm: ^^ This is how you cache it 😄

aarnphm:
> > I run into this issue when I forgot to close the terminal correct, so when trying to do `:q`, there is an error about a buffer still opening.
> > It turns out to be the unclosed toggleterm buffer.
> 
> Actually, non of above new toggleterm windows(lazygit, htop, python) can use `:q` to create an accidentally closed terminal.

Oh, because I have implement an on_open hook that bind q to close the current buffer.

aarnphm (approved):
LGTM

CharlesChiuGit:
@aarnphm Do u might port this cache mechanism to your own config? I still can't reproduce accidentally quit terminal after adding `q` to quit terminal.

aarnphm:
> @aarnphm Do u might port this cache mechanism to your own config? I still can't reproduce accidentally quit terminal after adding `q` to quit terminal. Or u just tested it?

Currently testing atm.

aarnphm:
Can confirm caching works.

Jint-lzxy:
Sorry for the delay. I'm making some minor modifications.

lua/keymap/completion.lua
```diff
@@ -0,0 +1,67 @@
...
+		end)
+		:with_noremap()
+		:with_silent()
+		:with_desc("lsp: Goto implementation"),
+	["n|gT"] = map_callback(function()
+			vim.lsp.buf.type_definition()
+		end)
+		:with_noremap()
+		:with_silent()
+		:with_desc("lsp: Goto type_definition"),
```
Jint-lzxy, line 64: @CharlesChiuGit I would prefer removing these keymaps as they will interfere with `lspsaga`'s `definitions.lua`. _(It will change the state of the servers in some cases, but `lspsaga` obviously wont know this)_

IMO It'll be better to request these features in the upstream repo.

lua/modules/configs/tool/telescope.lua
```diff
@@ -74,4 +74,18 @@ return function()
...
+				return not string.find(lhs, "Þ")
+			end,
+			layout_config = {
+				width = 0.6,
+				height = 0.6,
+				prompt_position = "top",
+			},
+		}
+		require("telescope.builtin").keymaps(opts)
+	end
```
Jint-lzxy, line 90: These functions can only be used after the plugin is loaded, but we should be able to automatically `require` the plugins after invoking the keymap.

btw I don't think these functions are globally accessable, did I miss something here?

Jint-lzxy (changes requested):

lua/keymap/completion.lua
CharlesChiuGit: Wow, doesn't know that. Will remove.

lua/modules/configs/tool/telescope.lua
CharlesChiuGit: ys, this function is globally accessible after this file is loaded. That's why I add `VeryLazy` to telescope's event.

Jint-lzxy: > we should be able to automatically `require` the plugins after invoking the keymap. This is what `lazy.nvim` does.

@CharlesChiuGit But not lazy-loading telescope will significantly increase startup time _(before UIEnter)_

Jint-lzxy: https://github.com/ayamir/nvimdots/pull/495/commits/9dfc7d7d5c079b57b587273138bef52333aa8cee should fix this.

CharlesChiuGit:
LGTM

Jint-lzxy (approved):
LGTM 🍺

@CharlesChiuGit New keymaps (if any) might need to be added to the wiki 👍

CharlesChiuGit:
The only newly added keymaps are those lsp ones, which is removed now. So I think here's no new keymaps.

PR #498, opened 2023-02-09
Added formatter for cuda

BinhaoQin:
Added formatter `clangfmt` for CUDA.


CharlesChiuGit:
Hmmm, I don't think it's a gd idea to add this. Since only very few ppl write cuda code.
Personally, I think it would be better to keep it for yourself.

PR #502, opened 2023-02-10
feat: Use CursorHold* to improve startup time

Jint-lzxy:
As mentioned in #501.

### ❗❗ The most obvious defect is this commit will cause external measuring tools to calculate the startup time incorrectly _(the delay of 200ms set by `updatetime` while loading is still within the acceptable range)_.

However, the "user-perceived" startup time _(especially in the case of directly opening a file for editing)_ will be **_greatly_** reduced (Tested on macOS)

@CharlesChiuGit Can you help me double-check whether there're feature breakages?

### DIFF
--- a/lua/modules/plugins/tool.lua
+++ b/lua/modules/plugins/tool.lua
@@ -6,7 +6,8 @@ tool["tpope/vim-fugitive"] = {
 }
 -- Please don't remove which-key.nvim otherwise you need to set timeoutlen=300 at `lua/core/options.lua`
 tool["folke/which-key.nvim"] = {
-	event = "VeryLazy",
+	lazy = true,
+	event = { "CursorHold", "CursorHoldI" },
 	config = require("tool.which-key"),
 }
 -- only for fcitx5 user who uses non-English language during coding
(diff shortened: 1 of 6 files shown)


CharlesChiuGit:
Sure, will check later the night.

ayamir:
The startuptime is too fast to perceive the improvement.

lua/modules/plugins/tool.lua
```diff
@@ -6,7 +6,8 @@ tool["tpope/vim-fugitive"] = {
 }
 -- Please don't remove which-key.nvim otherwise you need to set timeoutlen=300 at `lua/core/options.lua`
 tool["folke/which-key.nvim"] = {
-	event = "VeryLazy",
+	lazy = true,
+	event = { "CursorHold", "CursorHoldI" },
```
aarnphm, line 10: This will break atm.

Jint-lzxy: Ah yes, thanks for pointing that out 👍 Forgot that `timeoutlen` in this repo is `0`.

aarnphm (approved):

PR #505, opened 2023-02-11
events: quitting file type with q per pattern

aarnphm:

### DIFF
--- a/lua/core/event.lua
+++ b/lua/core/event.lua
@@ -28,6 +28,29 @@ vim.api.nvim_create_autocmd("BufEnter", {
 	end,
 })
 
+-- auto close some filetype with <q>
+vim.api.nvim_create_autocmd("FileType", {
+	pattern = {
+		"qf",
+		"help",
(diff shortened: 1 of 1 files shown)


lua/core/event.lua
```diff
@@ -28,6 +28,29 @@ vim.api.nvim_create_autocmd("BufEnter", {
...
+		"prompt",
+		"toggleterm",
+		"copilot",
+		"startuptime",
+		"tsplayground",
+		"PlenaryTestPopup",
+	},
+	callback = function(event)
+		vim.bo[event.buf].buflisted = false
+		vim.keymap.set("n", "q", "<CMD>close<CR>", { buffer = event.buf, silent = true })
```
CharlesChiuGit, line 50: `vim.api.nvim_set_keymap` is faster than `vim.keymap.set`

CharlesChiuGit (changes requested):

CharlesChiuGit (approved):

PR #506, opened 2023-02-11
chore(treesitter): only TSUpdate when UI is available

aarnphm:
Only setup treesitter if UI is available.

This will help make nvim --headless and running on CI faster


lua/modules/plugins/editor.lua
```diff
@@ -67,7 +67,11 @@ editor["romainl/vim-cool"] = {
 ----------------------------------------------------------------------
 editor["nvim-treesitter/nvim-treesitter"] = {
 	lazy = true,
-	build = ":TSUpdate",
+	build = function()
+		if #vim.api.nvim_list_uis() ~= 0 then
+			vim.cmd("TSUpdate")
```
CharlesChiuGit, line 72: `vim.api.nvim_command` is faster than `vim.cmd` for now.

CharlesChiuGit (changes requested):

CharlesChiuGit (approved):

PR #507, opened 2023-02-11
fix: use `vim.keymap.set` for filetype

aarnphm:
Fix a regression from 58b4832

### DIFF
--- a/lua/core/event.lua
+++ b/lua/core/event.lua
@@ -34,6 +34,7 @@ vim.api.nvim_create_autocmd("FileType", {
 		"qf",
 		"help",
 		"man",
+		"notify",
 		"nofile",
 		"lspinfo",
 		"terminal",
@@ -46,7 +47,7 @@ vim.api.nvim_create_autocmd("FileType", {
 	},
 	callback = function(event)
 		vim.bo[event.buf].buflisted = false
-		vim.api.nvim_set_keymap("n", "q", "<CMD>close<CR>", { buffer = event.buf, silent = true })
+		vim.keymap.set("n", "q", "<CMD>close<CR>", { buffer = event.buf, silent = true })
 	end,
 })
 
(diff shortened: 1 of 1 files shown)


CharlesChiuGit:
Maybe request review from Jint?

lua/core/event.lua
```diff
@@ -47,7 +47,7 @@ vim.api.nvim_create_autocmd("FileType", {
 	},
 	callback = function(event)
 		vim.bo[event.buf].buflisted = false
-		vim.api.nvim_set_keymap("n", "q", "<CMD>close<CR>", { buffer = event.buf, silent = true })
+		vim.keymap.set("n", "q", "<CMD>close<CR>", { buffer = event.buf, silent = true })
```
Jint-lzxy, line 50:
```console
:h nvim_buf_set_keymap

                                                       *nvim_buf_set_keymap()*
nvim_buf_set_keymap({buffer}, {mode}, {lhs}, {rhs}, {*opts})
    Sets a buffer-local |mapping| for the given mode.

    Parameters:  
      • {buffer}  Buffer handle, or 0 for current buffer

    See also:  
        |nvim_set_keymap()|
```

Also `alpha` has filetype 'nofile', pressing `q` inside the dashboard will immediately close it / throw error if it is the last window.

Sorry for the rush review, a kind of busy here

Jint-lzxy (changes requested):

lua/core/event.lua
aarnphm: no worries. This shouldn't be the main focus 😄

aarnphm: I also tried nvim_buf_set_keymap but it also raise error.

aarnphm, line 50:
```suggestion
		vim.api.nvim_buf_set_keymap(event.buf, "n", "q", "<CMD>close<CR>", { silent = true })
```

Jint-lzxy (approved):

Issue #504, opened 2023-02-11
migration: null-ls


aarnphm:
### Feature description

https://github.com/creativenull/efmls-configs-nvim hasn't seen development for a while now.

Also, I think that null-ls have better integration with mason.

aarnphm:
If we decided to migrate to null we will probably need to update the formatted logic as well.

CharlesChiuGit:
I'm working on it! Wait for my PR!

CharlesChiuGit:
Also this had been discussed in #500

aarnphm:
Got it. I will let you do it then.

aarnphm:
This has been merged in to main and will be closing now.

PR #513, opened 2023-02-13
migrate: from `efm` to `null-ls` ( #500 #504 )

CharlesChiuGit:
~~WIP~~


CharlesChiuGit:
Do we need to re-implement `FormatToggle` autocmd?

lua/modules/configs/completion/mason-tool-installer.lua
```diff
@@ -6,16 +6,18 @@ return function()
 			-- you can turn off/on auto_update per tool
 			-- "editorconfig-checker",
 
-			"stylua",
-
+			-- formatter
```
aarnphm, line 9: quick note on mason-tool-installer. Preferably I like the combination of `mason-null-ls` + null-ls and `mason-lspconfig` + nvim-lspconfig, as these two pairs provide a perfect solution for installing tools and setup their configuration.

Which means we can probably remove this plugin altogether.

CharlesChiuGit: That sound reasonable, since `mason-tool-installer` only install formatters for `null-ls` now. Will look into that now.

CharlesChiuGit: @aarnphm It seems https://github.com/jay-babu/mason-null-ls.nvim has a much shorter list of supported sources comparing to `mason-tool-installer`. I think I'll stick to `mason-tool-installer` for now.

aarnphm: mason-null-ls has a options to follow null-ls.nvim. If you already setup all of the sources for null-ls, then it will automatically install for you. In this way you don't have to context-switch and manage same dependencies at two places.

CharlesChiuGit: Ok, it seems `mason-null-ls` has some kind of auto-conversion, so it can install more sources than what's has been stated in it's source table: https://github.com/jay-babu/mason-null-ls.nvim#available-null-ls-sources

lua/modules/configs/completion/null-ls.lua
```diff
@@ -0,0 +1,71 @@
...
+		return builtin.with({
+			diagnostics_format = "#{m} [#{c}]",
+		})
+	end
+
+	local sources = {
+		-- formatting
+		b.formatting.black.with({ extra_args = { "--fast" } }),
+		b.formatting.prettierd.with({
+			extra_args = { "--no-semi", "--single-quote", "--jsx-single-quote" },
```
ayamir, line 21: No need to add this, because it will respect `.prettierrc.json` automatically.

CharlesChiuGit:
@aarnphm PTAL

I think it's better to config sources in `null-ls` since some formatters can also be used as lsp, like `markdownlint`. 
I'm not sure what's be setup via `mason-null-ls`' `automatic_setup`, so I think it's better to setup the sources by ourselves separately.

aarnphm (approved):
LGTM. cc @Jint-lzxy when you have bandwidth.

ayamir:
`FormatToggle` and `FormatterToggle` haven't been implemented.

ayamir (changes requested):
`:FormatToggle` and `:FormatterToggle` haven't been implemented.

CharlesChiuGit:
> `:FormatToggle` and `:FormatterToggle` haven't been implemented.

Ok, I'm on it.

CharlesChiuGit:
```lua
btn.formatting.clang_format.with({ command = "clang-format -style='{btnasedOnStyle: LLVM, IndentWidth: 4}'" }),
```
I found this line seems to have an error when running `:cheakhealth`.
`- ERROR clang_format: the command "clang-format -style='{btnasedOnStyle: LLVM, IndentWidth: 4}'" is not executable.`
But I'm not familiar with `clang-fornmat`, can someone help me with that?

Here's how `clang-format` configs in `null-ls`:
https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/formatting/clang_format.lua

ayamir:
> ```lua
> btn.formatting.clang_format.with({ command = "clang-format -style='{btnasedOnStyle: LLVM, IndentWidth: 4}'" }),
> ```
> 
> I found this line seems to have an error when running `:cheakhealth`. `- ERROR clang_format: the command "clang-format -style='{btnasedOnStyle: LLVM, IndentWidth: 4}'" is not executable.` But I'm not familiar with `clang-fornmat`, can someone help me with that?
> 
> Here's how `clang-format` configs in `null-ls`: https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/formatting/clang_format.lua
> 
> Maybe use
> 
> ```lua
> btn.formatting.clang_format.with({ args = "-style='{btnasedOnStyle: LLVM, IndentWidth: 4}'" }),
> ```
> 
> instead?

Just because typo of `btnasedOnStyle`, it should be `basedOnStyle` instead.

CharlesChiuGit:
> Just due to the typo of `btnasedOnStyle`, it should be `BasedOnStyle` instead.

Hmmm, still the same.
`- ERROR clang_format: the command "clang-format -style='{BasedOnStyle: LLVM, IndentWidth: 4}'" is not executable.`

CharlesChiuGit:
Do I need to install llvm or lldb or actucally use `clang-format` with `BasedOnStyle: LLVM`?

ayamir:
No need. But I can't see the ERROR after `:checkhealth`. Anyway, maybe there is no need to set it with these argus. Just keep it default also is fine.

CharlesChiuGit:
got it!

CharlesChiuGit:
@ayamir Have we tried https://github.com/lukas-reineke/lsp-format.nvim before?
It seems pretty handy.

CharlesChiuGit:
It seems like the old `formmatting.lua` is more powerful than `ls-format.nvim`. I think I'll tack it back and modify a bit.

ayamir:
Yes, but we choose to realize the functions by ourselves later.

CharlesChiuGit:
Do we need to keep the choice to use the actual LSP's formatting ability instead of only use `null-ls` from now on?

CharlesChiuGit:
In my experiment, `block_list` doesn't seem valid. How do I block a filetype for formatting?

Jint-lzxy:
> In my experiment, `block_list` doesn't seem valid. How do I block a filetype for formatting?（it doesn't seem to be included in the wiki?

@CharlesChiuGit Yep I plan to review & reimplement `:FormatToggle` and `:FormatterToggle` by the end of this week 👍 Some interaction logic has changed during this migration.

> Do we need to keep the choice to use the actual LSP's formatting ability instead of only use `null-ls` from now on? From a end-user view point, I think it's much easier to only set formatters in `null-ls` config rather than modify `formatting.lua`.
> 
> How do u guys think?

Not sure if I understood this question correctly, but IMHO this choice should be left to users. Some individuals' / organizations' projects would prefer to have formatting support provided by the language server _(especially if they already have a custom configuration file in use)_.

FWIW what we can do is remind users of those conflicting format approaches and ask them to confirm their preferences when encountering this for the first time. We can store these results in a cache file, so that the changes made will be persistent. _(This is also the feature that will be implemented when refactoring `:FormatterToggle`)_

CharlesChiuGit:
@Jint-lzxy Then i'll just change efm to null-ls and leave everything else untouched. @ayamir how do u think?

CharlesChiuGit:
> `:FormatToggle` and `:FormatterToggle` haven't been implemented.

@ayamir PTAL

aarnphm (approved):

CharlesChiuGit:
side note: I think some of `formatting.lua`'s setting should be prot to `core/settings.lua`. But maybe @Jint-lzxy has already take that into consideration lol

CharlesChiuGit:
[`2ff5d96` (#513)](https://github.com/ayamir/nvimdots/pull/513/commits/2ff5d9655f8fbede7e6745e5765c575b70c058c4) moves `lsp` and `null-ls` to `core/settings/lua`. Will add this to wiki once merged.

ayamir (approved):
LGTM 👍

aarnphm:
Feel free to merge it in when it is gtg

PR #514, opened 2023-02-13
feat: add/replace plugins ( #494 )

CharlesChiuGit:
Replace/add plugins for better performance or interesting features.
1. feat(paint.nvim): add highlights for type hints
2. feat(cmp-treesitter): add treesitter node for nvim-cmp source
3.  migrate: use `pysan3/fcitx5.nvim` instead of `zhaosheng-pan/vim-im-select`
4. feat(autoclose.nvim): three times faster than `windwp/nvim-autopairs`


CharlesChiuGit:
Other plugins that's might need a poll:
1. https://github.com/romainl/vim-cool -> https://github.com/asiryk/auto-hlsearch.nvim , but it seems the later is a bit slower.
2. https://github.com/jcdickinson/codeium.nvim, basically free copilot. But I haven't test it for a long run.
3. https://github.com/roobert/bufferline-cycle-windowless.nvim, skip past buffers that are already open in an existing window when doing `BufferCycle`

aarnphm:
Few things to note here:
1. paint.nvim should only be enabled when `show_bigfile` is set to true.
2. I don't see the boost for using autoclose.nvim (maybe because I don't use nvim-autopairs that extensively).
3. codeium.nvim is a nice add, but I think user should be able to do that. copilot.lua is ok for now.
4. vim-cool is after than auto-hlsearch.nvim from initial testing.

CharlesChiuGit:
> 1\. paint.nvim should only be enabled when `show_bigfile` is set to true.

I don't get it. Why should we enabled it when `show_bigfile = true`? It also works on small files fine, no?

> 2\. I don't see the boost for using autoclose.nvim (maybe because I don't use nvim-autopairs that extensively).

I check it's profile via `:Lazy`.

> 3\. codeium.nvim is a nice add, but I think user should be able to do that. copilot.lua is ok for now.

Nice! Let's see how ayamir and Jint think of it.

> 4\. vim-cool is after than auto-hlsearch.nvim from initial testing.

U mean "**faster** than"? Yes, that's also true on my machine. `vim-cool` takes about 0.3ms to load, and `auto-hlsearch.nvim` takes about 0.5ms. I just brought it out b/c it's written in pure lua.

aarnphm:
> 4. vim-cool is after than auto-hlsearch.nvim from initial testing.
yes a typo.
>  paint.nvim should only be enabled when show_bigfile is set to true
oh maybe I'm wrong. I thought that plugins are for larger file.

CharlesChiuGit:
> oh maybe I'm wrong. I thought that plugin is used for larger file.

Not necessarily. It just works better than `tree-sitter-comment` on larger files. Doesn't mean it should "only" works on larger files.

ayamir:
``

> Other plugins that's might need a poll:
> 
> 1. https://github.com/romainl/vim-cool -> https://github.com/asiryk/auto-hlsearch.nvim , but it seems the later is a bit slower.
> 2. https://github.com/jcdickinson/codeium.nvim, basically free copilot. But I haven't test it for a long run.
> 3. https://github.com/roobert/bufferline-cycle-windowless.nvim, skip past buffers that are already open in an existing window when doing `BufferCycle`

1. I can see any improvement than `vim-cool`.
2. AI powered engine maybe need to add as comment like what we do for `tabnine.nvim`.
3. No need to add this, because sometimes I just want to see current buffer.

ayamir:
What's more, `fcitx5.nvim` should keep as comment only for users who use `fcitx5`.

ayamir:
Other plugins no problems.

CharlesChiuGit:
> 2\. AI powered engine maybe need to add as comment like what we do for `tabnine.nvim`.

> `fcitx5.nvim` should keep as comment only for users who use `fcitx5`.

noted

CharlesChiuGit:
@ayamir PTAL

ayamir (approved):

Issue #426, opened 2023-01-16
新建文件时不显示statusline要如何解决？


nkta3m:
### Version confirmation

- [ ] Confirm

### Neovim version

nvim 0.82

### Operating system/version

Windows 10

### Terminal name/version

Alacritty

### Branch info

main (Default/Latest)

### Fetch Preferences

HTTPS (use_ssh = false)

### How to reproduce the issue

1. open nvim
2. space f n
3. save to test.py

### Expected behavior

正确显示

### Actual behavior

nvim test.py 或者nvim，快捷键 space-f-n,新建一个空文件，:save test.py，statusline都不能正确显示，请教是正常的吗？

Jint-lzxy:
This _is_ the correct _(and yet not that intuitive)_ behavior, because we haven't registered the `BufNewFile` event for `lualine` and its two dependencies. This will be resolved in #425.

nkta3m:
明白了，谢谢

nkta3m:
我又遇到另一个问题，新建md文件，出现如下提示：
![2023-01-16 13_08_04-Window](https://user-images.githubusercontent.com/13484700/212603054-924b7dc9-efbc-4279-be02-784d28d905a1.png)

Jint-lzxy:
This is also related to the `BufNewFile` event _(++the language server would also stop working under certain circumstances)_. Please wait until the migration (#422) is finished.

A temporary workaround is to create such files externally using shell commands like `touch`.

Jint-lzxy:
`:h BufNewFile`
```vimdoc
							*BufNewFile*
BufNewFile			When starting to edit a file that doesn't
				exist.  Can be used to read in a skeleton
				file.
```

ayamir:
The correct steps to use `<leader>fn` to create a new file and use opened buffer to edit are following:
1. `<leader>fn`: create a empty buffer.
2. `:w xxx.xx`: write to a file called `xxx.xx`.
3. `:e`: reload this buffer with the saved file.

Then you will see all of them works for fine including `lsp`, `lualine` and `bufferline`.

ayamir:
> 我又遇到另一个问题，新建md文件，进入insert后，出现如下提示：但是在test.py中不会出现
> 
> ![2023-01-16 13_08_04-Window](https://user-images.githubusercontent.com/13484700/212603054-924b7dc9-efbc-4279-be02-784d28d905a1.png)

[...]

PR #517, opened 2023-02-14
perf: only attach lualine on LSP setup

aarnphm:
Note that we are using lspsaga with lualine. This means it will be a bit more
expensive to load lspsaga with BufReadPost. We probably only need to use lualine
once LspAttach is called.

This makes it around 10ms faster startup time.

### DIFF
--- a/lua/modules/plugins/ui.lua
+++ b/lua/modules/plugins/ui.lua
@@ -36,7 +36,7 @@ ui["lukas-reineke/indent-blankline.nvim"] = {
 }
 ui["nvim-lualine/lualine.nvim"] = {
 	lazy = true,
-	event = { "BufReadPost", "BufAdd", "BufNewFile" },
+	event = "LspAttach",
 	config = require("ui.lualine"),
 }
 ui["zbirenbaum/neodim"] = {


Jint-lzxy (commented):
One downside is statusline won't be enabled until the user opens a filetype that has pre-configured server(s). _(e.g., statusline won't be enabled in `:enew`, `*.txt`, `*.log`)_

ayamir (commented):
We set it because https://github.com/ayamir/nvimdots/issues/426

CharlesChiuGit:
Either we can remove winbar from lualine but the windows will jitter a bit when LspStart, or keep it in current settings and a bit delay on startup. 

Personally I prefer the later.

aarnphm:
> One downside is statusline won't be enabled until the user opens a filetype that has pre-configured server(s). _(e.g., statusline won't be enabled in `:enew`, `*.txt`, `*.log`)_

You are correct. I just feel like attaching lspsaga at startup is relatively heavy.

ayamir:
We can utilize `exclude` array to prevent `lspsaga` required by `lualine`.

https://github.com/ayamir/nvimdots/blob/78ab09e7121bdef3ae76212faa2b380a6cf17fec/lua/modules/configs/ui/lualine.lua#L16-L22

`LspAttach` is triggerd after lsp client attached to the buffer. So maybe we can set `exclude` dynamically by checking which lsp servers were installed. But there are no ways to get lsp servers from `mason`. We can just get all of packages including `stylua` which are not lsp servers.

aarnphm:
> ```lua
> require("mason-registry").get_installed_package_names()
> ```

Have a bit of time thinking about this. This means we are explicitly depends on mason. idk if we should do that. If you want to do I can go ahead and implement the logic.

ayamir:
> > ```lua
> > require("mason-registry").get_installed_package_names()
> > ```
> 
> Have a bit of time thinking about this. This means we are explicitly depends on mason. idk if we should do that. If you want to do I can go ahead and implement the logic.

IMO no need to do this.

PR #524, opened 2023-02-16
fix: remove autocompletion on dap-repl

aarnphm:
This PR address an issue when after exiting the dap-repl buffer,
it mess up with the current completion system.

This has to do with us not using omni completion (which dap-repl.ext.autocmpl relies on).

Might be a good idea to write our own completion system for dap-repl if needed.


CharlesChiuGit:
Seems like a really big project? Or maybe it can be a plugin itself?

aarnphm:
i don't find dap-repl very useful.

CharlesChiuGit:
@aarnphm is this PR still been working on?

aarnphm:
nope. I have been busy for a while with work

ayamir:
@aarnphm Wish you all the best.

PR #525, opened 2023-02-16
refactor: tidy mason handler

aarnphm:
This PR tidy up mason handlers logics and add a new behaviour for configuration
where users can return a functions that takes in an options

an example can be found in clangd:

```lua
return function(options)
require("lspconfig").clangd.setup { ... }
end
```

This means users don't have to worry about overriding default lsp setup in
setup_handlers,

and the end users can just add a new server configuration file in their config.

For example, under lua/modules/config/completion/servers, create a new
configuration for elm

```lua
return {
...
}
```

User then can just install the LSP server with `:LspInstall elm` and it will
work as expected with all of the capabilities and on_attach handlers.


CharlesChiuGit:
After a quick view, I have one question. What if user didn't create an `servers/elm.lua`? Do we have a default config for it?

Specifically, I mean this part:

https://github.com/ayamir/nvimdots/blob/3dda404f5000e196ec5955754df34a1816c229e3/lua/modules/configs/completion/lsp.lua#L67-L72

CharlesChiuGit:
Also, I personally don't think status check is needed. It's better to diagnose if some plugin is broken.

CharlesChiuGit:
Since @Jint-lzxy is refactoring `formatting.lua`, I think this part 

https://github.com/aarnphm/ayamir-nvimdots/blob/e81dd04a688e8a359bae314c3dbab378e91e40b7/lua/modules/configs/completion/lsp.lua#L57-L60 

should be placed in `formatting.lua`

lua/modules/configs/completion/lsp.lua
```diff
@@ -40,71 +40,115 @@ return function()
...
+			local lspconfig = require("completion.servers")[lsp_name]
+			if type(lspconfig) == "function" then
+				--- NOTE: case where language server requires its own setup
+				--- Make sure to call require("lspconfig")[lsp_name].setup() in the function
+				--- See clangd.lua for example.
+				lspconfig(opts)
+			elseif type(lspconfig) == "table" then
+				nvim_lsp[lsp_name].setup(vim.tbl_deep_extend("force", opts, lspconfig))
+			else
+				error(
```
CharlesChiuGit, line 134: Will this called `nvim-notify`? Or it just shows in the cmd line?

CharlesChiuGit:
I do like the idea of config LSPs entirely in the `server/` folder!

lua/modules/configs/completion/lsp.lua
aarnphm: error respect vim.notify.

aarnphm:
> Since @Jint-lzxy is refactoring `formatting.lua`, I think this part
> 
> [aarnphm/ayamir-nvimdots@`e81dd04`/lua/modules/configs/completion/lsp.lua#L57-L60](https://github.com/aarnphm/ayamir-nvimdots/blob/e81dd04a688e8a359bae314c3dbab378e91e40b7/lua/modules/configs/completion/lsp.lua#L57-L60)
> 
> should be placed in `formatting.lua`
> 
> And some user might only want to using LSP formatting ability on certain LSP. We might need to deal with each server individually.

Yep, I will wait until everything else is done. Just want to put the PR here so that I don't forget about it 😄

aarnphm:
> I do like the idea of config LSPs entirely in the `server/` folder! Users can now just focus on `some_lsp.lua`.

if we agree on this change then I can update the Wiki afterward.

Jint-lzxy (commented):
Great improvement! I haven't seen any issues yet.

Will review this PR after https://github.com/ayamir/nvimdots/pull/513#issuecomment-1433064481 is implemented 👍

Jint-lzxy:
Changes will be pushed after supporting lspsaga@v0.2.7! Currently working on that :)

Jint-lzxy:
Next push will sort out these helper functions and put them in `utils.lua`

Jint-lzxy:
@CharlesChiuGit https://github.com/ayamir/nvimdots/pull/525/commits/032dd358ad7c68bc4fc9a128cee99294f39ca23a set up `null-ls` using another approach and fixed previously broken `clang_format` config. Can you review these changes? Thanks in advance :)

Jint-lzxy:
https://github.com/ayamir/nvimdots/pull/525/commits/e51cca4086614cddb4e24335d342dd96208ac697 fixed user commands related to lsp formatting. `:FormatToggle` will disable lsp formatting in the current buffer, whereas `:FormatterToggleFt` will disable lsp formatting in specific fts. This can also be configured with:
```lua
-- Filetypes in this list will skip lsp formatting if rhs is true
---@type table<string, boolean>
settings["formatter_block_list"] = {
	-- Example
	lua = false,
}
```

CharlesChiuGit:
Got an error when running this branch for the first time.
圖片

CharlesChiuGit:
@Jint-lzxy I think we should port this condition as a table and placed it in the `core/settings.lua`
```lua
elseif client.name ~= "lua_ls" and client.name ~= "tsserver" and client.name ~= "clangd" then
```

CharlesChiuGit:
> @CharlesChiuGit [032dd35](https://github.com/ayamir/nvimdots/commit/032dd358ad7c68bc4fc9a128cee99294f39ca23a) set up `null-ls` using another approach and fixed previously broken `clang_format` config. Can you review these changes? Thanks in advance :)

For those changes related to `null-ls` are fine, tho I would prefer use `configs/completion/null-ls.lua` and `configs/completion/formatters/` instead, as an analogy of `configs/completion/lsp.lua` and `configs/completion/servers/`.

CharlesChiuGit:
Also, is this all the refactor that will be applied to `formatting.kua`?

CharlesChiuGit:
> > @CharlesChiuGit [032dd35](https://github.com/ayamir/nvimdots/commit/032dd358ad7c68bc4fc9a128cee99294f39ca23a) set up `null-ls` using another approach and fixed previously broken `clang_format` config. Can you review these changes? Thanks in advance :)
> 
> For those changes related to `null-ls` are fine, tho I would prefer use `configs/completion/null-ls.lua` and `configs/completion/formatters/` instead, as an analogy of `configs/completion/lsp.lua` and `configs/completion/servers/`.

Actually, the reason why I use `mason-null-ls`' auto setup is to let user config `null-ls` in `core/settings.lua` if they only use default config. The approach in this PR will need the user to config sources in `null-ls/init.lua`.

Jint-lzxy:
> Got an error every time I run this branch.

Also noticed this. AFAIU there's a logic error while registering the servers _(the registry function returns a function but that is never called(?))_

Jint-lzxy:
> Got an error every time I run this branch.

Also noticed this. AFAIU there's a logic error while registering the servers _(the registry function returns a function but that function is never called(?))_

Jint-lzxy:
> @Jint-lzxy I think we should port this condition as a table and placed it in the `core/settings.lua`
> 
> ```lua
> elseif client.name ~= "lua_ls" and client.name ~= "tsserver" and client.name ~= "clangd" then
> ```

That's a great idea! Will do.

Jint-lzxy:
> Also, is this all the refactor that will be applied to `formatting.kua`?

I think so. That logic error has been fixed.

CharlesChiuGit:
> > @Jint-lzxy I think we should port this condition as a table and placed it in the `core/settings.lua`
> > ```lua
> > elseif client.name ~= "lua_ls" and client.name ~= "tsserver" and client.name ~= "clangd" then
> > ```
> 
> That's a great idea! Will do.

I think this part should be also be moeved to `formatting.lua`'s filter.
https://github.com/aarnphm/ayamir-nvimdots/blob/bb15376ecf23a0d40ce63eb54b923ecf0b020818/lua/modules/configs/completion/lsp.lua#L47-L50

I can do that if you're busy.

Jint-lzxy:
> Actually, the reason why I use `mason-null-ls`' auto setup is to let user config `null-ls` in `core/settings.lua` if they only use default config. The approach in this PR will need the user to config sources in `null-ls/init.lua`.

What I considered is this approach will make _custom_ configuration more difficult _(and the server(s) will be setup twice)_ as this is not the method provided in null-ls's "basic" documentation, which may lead to further confusion. Maybe we need to discuss this a bit more 😄

CharlesChiuGit:
> What I considered is this approach will make _custom_ configuration more difficult _(and the server(s) will be setup twice)_ as this is not the method provided in null-ls's "basic" documentation, which may lead to further confusion. Maybe we need to discuss this a bit more 😄

Sure! We can discuss on this after other things are settled.

Jint-lzxy:
> I can do that if you're busy.

It's okay 😃 I'm rewriting part of this feature, and it may take some time :) These trivial updates will also be included.

Jint-lzxy:
https://github.com/ayamir/nvimdots/pull/525/commits/6400bb723b73e936f635d8f0af1d3a7f2de629c4 fixed those major issues, including:
- We cannot directly `require` [a folder](https://github.com/aarnphm/ayamir-nvimdots/blob/refactor/mason-handlers/lua/modules/configs/completion/lsp.lua#L114)
- `use_server_formatting_provider` was removed, this will be configured in `settings.lua` instead
- Simplified code logic and removed unnecessary functions to make `lsp.lua` easier to maintain IMO

cc @aarnphm @CharlesChiuGit

Jint-lzxy:
https://github.com/ayamir/nvimdots/pull/525/commits/9b1be95de132e598719cfd8e2d8dd773c0ca312b implemented https://github.com/ayamir/nvimdots/pull/525#issuecomment-1436212816.

aarnphm (commented):
👍 

Just realized I can't accept my own PR 😄

lua/modules/configs/completion/formatting.lua
```diff
@@ -105,7 +84,7 @@ function M.format_filter(clients)
 		end)
 		if status_ok and formatting_supported and client.name == "null-ls" then
 			return "null-ls"
-		elseif client.name ~= "lua_ls" and client.name ~= "tsserver" and client.name ~= "clangd" then
+		elseif not require("core.settings").server_formatting_block_list[client.name] then
 			return status_ok and formatting_supported and client.name
```
aarnphm, line 88:
```suggestion
		elseif not require("core.settings").server_formatting_block_list[client.name] and status_ok and formatting_supported then
			return client.name
```

minor cleanup @Jint-lzxy

Jint-lzxy:
LGTM :) @CharlesChiuGit We may discuss https://github.com/ayamir/nvimdots/pull/525#issuecomment-1436835516 now 👍 Everything else has been settled.

lua/modules/configs/completion/lsp.lua
```diff
@@ -56,57 +54,56 @@ return function()
...
-		end,
-
-		lua_ls = function()
-			local _opts = require("completion.servers.lua_ls")
-			local final_opts = vim.tbl_deep_extend("keep", _opts, opts)
-			nvim_lsp.lua_ls.setup(final_opts)
-		end,
-	})
+	-- Set lsps that are not supported by `mason.nvim` but supported by `nvim-lspconfig` here.
+	if vim.fn.executable("html-languageserver") then
```
CharlesChiuGit, line 104: Why is this section needed?
`html` is supported by mason, no?

CharlesChiuGit: Or this is a demo for those unsupported lsp?

Jint-lzxy: > Why is this section needed? `html` is supported by mason, no?

AFAIK, no. Mason only provides `html-lsp` but not `html-languageserver` used by vscode. This is the exact reason why we setup `vscode-html-languageserver` separately before. See [Wiki: Prerequisites#tools-for-plugins](https://github.com/ayamir/nvimdots/wiki/Prerequisites#tools-for-plugins).

Jint-lzxy: > Or this is a demo for those unsupported lsp?

Yep this can also be used as a demo 😄

CharlesChiuGit: Actually, they are the same.
mason side: https://github.com/williamboman/mason.nvim/blob/main/lua/mason-registry/index/html-lsp/init.lua
nvim-lspconfig side: https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/html.lua
mason-lspconfig mapping: https://github.com/williamboman/mason-lspconfig.nvim/blob/3f9c0ea157e88b08e2d11ec65002eeb6e421f0da/lua/mason-lspconfig/mappings/server.lua#L64

CharlesChiuGit: And that's why I put html to mason's `ensure-install` table lol

aarnphm: we can add in the dartls as example as well.

CharlesChiuGit: > we can add in the dartls as example as well.

ys, that's what I'm about to say lol

CharlesChiuGit: > Actually, they are the same. mason side: [williamboman/mason.nvim@`main`/lua/mason-registry/index/html-lsp/init.lua](https://github.com/williamboman/mason.nvim/blob/main/lua/mason-registry/index/html-lsp/init.lua?rgh-link-date=2023-02-26T07%3A20%3A22Z) nvim-lspconfig side: [neovim/nvim-lspconfig@`master`/lua/lspconfig/server_configurations/html.lua](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/html.lua?rgh-link-date=2023-02-26T07%3A20%3A22Z) mason-lspconfig mapping: [williamboman/mason-lspconfig.nvim@`3f9c0ea`/lua/mason-lspconfig/mappings/server.lua#L64](https://github.com/williamboman/mason-lspconfig.nvim/blob/3f9c0ea157e88b08e2d11ec65002eeb6e421f0da/lua/mason-lspconfig/mappings/server.lua#L64)

They're installing the same npm pkg, and using the same bin name.

Jint-lzxy: Ah thanks for the update 😆 Looks like I missed some commits for `mason`

> we can add in the dartls as example as well.

Sure! Will do.

Jint-lzxy: https://github.com/ayamir/nvimdots/pull/525/commits/b5ffe179e3e554f66ec0e30856a26d3c4287ea06

CharlesChiuGit:
The reason why I would prefer the auto setup form `mason-null-ls` is mostly that users only need to set your non-LSP sources in `core/settings.lua`, without needing to edit `configs/completion/null-ls.lua` if they only the default null-ls settings.

Thought this is not the official way to setup non-LSP source with extra settings in `null-ls`, but it's still pretty much the same and calling `null-ls` registry under the hood.

For those user who would like to use extra settings for non-LSP sources, I don't think it's super difficult for them to check out how to state them in `mason-null-ls`, we just need to add a note in `null-ls.lua` or wiki.

Overall, I think it's much simpler to use `mason-null-ls`' auto setup.

Jint-lzxy:
@CharlesChiuGit I'm concerned about some corner cases _(i.e., servers that cannot be installed via `mason`, such as `dart_format`)_. If we adopt that convention, users will almost _surely_ add `dart_format` to `null_ls_deps`, and then find out this has no effect - the server is not installed, and document formatting is not activated. FWIW this will be quite frustrating for users who are not familiar with `neovim` as they have very few documentations to refer to _(standard configs are not as they are now)_. Apart from this, everything LGTM 👍

Jint-lzxy:
IMO it would be better to have a solution that is "universally available" 😄

CharlesChiuGit:
@Jint-lzxy 
To the best of my knowledge, `mason` don't include this kind of sources, like `dartls` and `dart_format`, b/c their already included in the `dart` runtime pkgs. So if a dart user already have `dart` runtime installed, then adding `dart_format` in `core/settings.lua`'s `null_ls_deps` will certainly take effect in the scoop of `null-ls`.

Jint-lzxy:
@CharlesChiuGit https://github.com/ayamir/nvimdots/pull/525/commits/96b6d5780ab28f8392623aa15e07b76e09051f2d. Hope I correctly understood what you mean 😄

Jint-lzxy:
This commit also set lazy-loading for null-ls to reduce initialization time.

CharlesChiuGit:
Mostly right! But I'll modify it a bit.

CharlesChiuGit:
LGTM!

CharlesChiuGit (approved):

Jint-lzxy:
@CharlesChiuGit Hmm I was deliberately avoiding this approach as it complicates the initialization process _(compared with the method provided by null-ls's docs)_. Using `setup_handlers` will setup the servers specified in that function twice, so I just let null-ls complete such initilization. (`memoize` will check whether the server [is already registered](https://github.com/jay-babu/mason-null-ls.nvim/blob/main/lua/mason-null-ls/automatic_setup.lua#L7), but overwriting the configs using setup_handlers will not)

Let me know if I misunderstood something 🍰

Jint-lzxy:
To be short, as we have already setup the servers that require additional configuration in null-ls, mason-null-ls will not set up these servers again.

CharlesChiuGit:
Aha, got it. If that's the case then it's probably better to revert my commit.
But I have this commit based on [the discussion](https://github.com/jay-babu/mason-null-ls.nvim/discussions/51) I had with the plugin author.

Jint-lzxy:
The plugin author wrote:
> Hmmm, I am not actually sure what happens when `null-ls.register` is called for the same exact source.

This is the cause of such confusion 😄

CharlesChiuGit:
Got it! Thanks!

CharlesChiuGit (approved):

Jint-lzxy (approved):
🍦

CharlesChiuGit:
https://github.com/ayamir/nvimdots/wiki/Usage#modify-lsps-linters-and-formatters

Sorry for the late update. PTAL.

aarnphm:
LGTM. I haven't use the structure for a while now so need a quick refresh 😄

PR #532, opened 2023-02-18
feat: unify icons with cod icons

huge-pancake:
Unify icons with cod icons


aarnphm (changes requested):
I think if user wish to change this, user can. I'm a bit hesitant to merge this upstream. Maybe you can have an options for people to switch different icon type backend.

CharlesChiuGit:
codicon is not normal user will use, u can use it. But i'm leaning towards close this pr.

Jint-lzxy:
Sorry about this, but this PR will be closed as users can customize the icon set themselves, and unexpected changes may occur if we merge this upstream.

But anyway, thank you for the contribution 👍

PR #540, opened 2023-02-25
Update editor.lua

AngelontheRoad:
fix treehopper keymap


Jint-lzxy (approved):
Good catch! Thanks for the contribution 🍨

PR #555, opened 2023-03-07
feat(lsp): use `pcall` for error handling

Jint-lzxy:
Just realized wrapping `require` inside `pcall` would do the trick 😄


CharlesChiuGit:
That's genius! 
LGTM

aarnphm:
Welp i definitely forgot about pcall 😄

PR #571, opened 2023-03-15
feat(ci): update lockfile only when there are new commits

Jint-lzxy:
This PR would stop `update_lockfile.yml` from messing up the commit history.


Jint-lzxy:
Possible outcome: [Jint-lzxy/nvimconfig#81](https://github.com/Jint-lzxy/nvimconfig/actions/runs/4427122305/jobs/7764281775)

PR #575, opened 2023-03-16
fix(script-win): initialize user functions correctly

Jint-lzxy:
Turns out that `Prompt` is a [pwsh built-in](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_prompts) 😢 This PR also unifies function names for all platforms.

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -5,6 +5,8 @@
 
 Set-StrictMode -Version 3.0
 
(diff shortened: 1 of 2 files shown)


CharlesChiuGit:
will test tomorrow.

aarnphm:
I'm not on Windows so I can't help with this 😄

CharlesChiuGit (approved):
LGTM

PR #598, opened 2023-03-25
pref(cmp)!: update keymap for `cmp.close`

Jint-lzxy:
This commit changes the keymap for `cmp.close` from `<C-e>` to `<C-w>`.

This way:
* Makes the keymap more intuitive and easier to memorize (`<C-w>` --> Close the focused window on most platforms)
* Makes `cmp.close` easier to type _(`w` is closer to `<CTRL>` than `e`)_

As this change does not conflict with `<C-w>` in normal mode, no conflicting keymap was found for the time being.

### DIFF
--- a/dots.tutor
+++ b/dots.tutor
@@ -105,7 +105,7 @@ The completion window will always show no matter what file you are editing.
 You can use `<Tab>`{normal} and `<S-Tab>`{normal} to select next and previous candidate in
 the completion window and use `<CR>`{normal} to confirm completion. You can also use
 `<C-n>`{normal} and `<C-p>`{normal} and `<C-y>`{normal} to do it. Sometimes you don't want confirm
-completion but just want to new a line, you can use `<C-e>`{normal} to close the
+completion but just want to new a line, you can use `<C-w>`{normal} to close the
 window manually.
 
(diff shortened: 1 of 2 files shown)


ayamir:
No problem, but this line also need to be modified if so.
https://github.com/ayamir/nvimdots/blob/441321b81d36b5fb91a6cf62f00b517372e3ffbe/dots.tutor#L108

Jint-lzxy:
https://github.com/ayamir/nvimdots/pull/598/commits/109881e4cdde2d851933f050c6cb60c08b96367f

ayamir (approved):

kang8:
By the way, the default behavior of `<C-w>` is to delete the word before the cursor in neovim, see [`:h i_CTRL_w`](https://neovim.io/doc/user/insert.html#i_CTRL-W). This may have an impact on some users.

Also, in nvim-cmp, if `<C-e>` is not overridden, its default behavior is to call [`abort()`](https://github.com/hrsh7th/nvim-cmp/blob/93cf84f7deb2bdb640ffbb1d2f8d6d412a7aa558/lua/cmp/config/mapping.lua#L38-L40).

PR #600, opened 2023-03-27
perf(icon): add cmp icons for `codeium` and `tabnine`

charliie-dev:
Without these icons, `cmp` will throw errors when showing suggestion windows.
No very creative tho lol

### DIFF
--- a/lua/modules/utils/icons.lua
+++ b/lua/modules/utils/icons.lua
@@ -160,8 +160,8 @@ local data = {
 		Vim = "",
 	},
 	cmp = {
-		Codeium = "",
-		TabNine = "",
+		Codeium = "",
+		TabNine = "",
 		Copilot = "",
 		Copilot_alt = "",
 		nvim_lsp = "",


Jint-lzxy:
Perhaps [`\uf1ab`](https://fontawesome.com/v4/icon/language) is a better alternative?

CharlesChiuGit:
> Perhaps [`\uf1ab`](https://fontawesome.com/v4/icon/language) is a better alternative?

sure!

CharlesChiuGit:
But `nvim_lsp` is using the same icon?

lua/modules/utils/icons.lua
```diff
@@ -160,6 +160,8 @@ local data = {
 		Vim = "",
 	},
 	cmp = {
+		Codeium = "",
```
ayamir, line 163: this icon is from https://github.com/jcdickinson/codeium.nvim.

CharlesChiuGit: ys

CharlesChiuGit:
> IMO new icons are needed. They usually give completion by sentence or paragraph, which is different from traditional lsp client's behavior. Maybe `nf-cod-symbol_keyword` in nerd font is a good choice?

That makes a lot of sense. Updated!

PR #604, opened 2023-03-31
fix: easyalign keymap

fecet:
The original easyalign keymap use `<Plug>(EasyAlign)` , which will not correctly hook lazyloading of easyalign as that's not actually us cmd `EasyAlign`
https://github.com/ayamir/nvimdots/blob/c56cd604dc3b9f7149b89f245206daf0b00b369c/lua/modules/plugins/editor.lua#L48

### DIFF
--- a/lua/keymap/editor.lua
+++ b/lua/keymap/editor.lua
@@ -69,11 +69,7 @@ local plug_map = {
 	["n|<leader><leader>D"] = map_cr("DiffviewClose"):with_silent():with_noremap():with_desc("git: Close diff"),
 
 	-- Plugin: vim-easy-align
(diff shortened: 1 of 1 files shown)


aarnphm (approved):
LGTM. cc @CharlesChiuGit since I don't use EasyAlign.

ayamir (approved):

CharlesChiuGit:
> LGTM. cc @CharlesChiuGit since I don't use EasyAlign.

Actually I don't use EazyAlign either lol.

fecet:
I suppose I wouldn't use it either... I was only testing it to verify that my migration didn't cause any issues.

PR #607, opened 2023-04-01
feat: use hop in visual mode

fecet:
An important feature (to me) of hop.nvim is it support be used in visual mode:
```
Use it with commands like v, d, c, y to visually select/delete/change/yank up to your new cursor position.
```
But the current map only support normal mode.

Weird thing is simply changing it to
```
["nv|<leader>w"] = map_cu("HopWord"):with_noremap():with_desc("jump: Goto word")
```
does not work.


ayamir (approved):

ayamir:
https://github.com/ayamir/nvimdots/blob/505b25f97f9c2a2ea9e634a8c269360a6f40c528/lua/keymap/bind.lua#L55

PR #611, opened 2023-04-02
fix(cmp): handle LuaSnip jump failures

Jint-lzxy:
This PR should finally address https://github.com/ayamir/nvimdots/discussions/599.

Looks like LuaSnip regards the situation of "cursor being inside a snippet but unable to jump forward" as `locally_jumpable`. This PR checks whether the jump was successful by determining cursor positions before and after calling LuaSnip.

Here, `vim.schedule_wrap` is used to avoid calling this function before `vim.fn.feedkeys` is evaluated _(`feedkeys()` does not block the main thread)_.

### DIFF
--- a/lua/modules/configs/completion/cmp.lua
+++ b/lua/modules/configs/completion/cmp.lua
@@ -21,6 +21,17 @@ return function()
 		}
 	end
 
+	---Handling situations where LuaSnip failed to perform any jumps
+	---@param r integer @Cursor position (row) before calling LuaSnip
+	---@param c integer @Cursor position (col) before calling LuaSnip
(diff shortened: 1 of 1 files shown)


ayamir (approved):
LGTM 👍

huge-pancake:
image

It seems that `unpack` is deprecated.

huge-pancake:
`unpack` now is `table.unpack` since Lua 5.2

CharlesChiuGit:
> image
> 
> 
> 
> It seems that `unpack` is deprecated.

unpack is deprecated due to the lua version embedded in neovim. We can't do much about it. Better just ignore it lol

PR #633, opened 2023-04-08
feat!: add initial support for lsp semantic tokens

Jint-lzxy:
This PR added initial support for lsp semantic tokens. It only implements **a small part** of semantic tokens.

Further development would require testing on more supported languages.


CharlesChiuGit (approved):
LGTM

PR #648, opened 2023-04-11
chore(kind_text): cleanup

Jint-lzxy:


CharlesChiuGit (approved):
LGTM!

lua/modules/configs/completion/cmp.lua
```diff
@@ -41,18 +41,14 @@ return function()
...
-			opts = {}
-		end
+	local function cmp_format(opts)
+		opts = opts or {}
 
 		return function(entry, vim_item)
-			if opts.before then
-				vim_item = opts.before(entry, vim_item)
-			end
+			vim_item = opts.before and opts.before(entry, vim_item)
```
kagaya85, line 48: In my usage scenario, I found this change will cause the `vim_item` to have a null value

CharlesChiuGit: do u have a screenshot of the error and the null cmp item?

CharlesChiuGit: but from the logic, this should perform the same as the previous code.

kagaya85: yeah, it just looks like this:
ScreenShot_2023-04-12_at_14 11 45@2x

This error may occur when the `opts.before` is `nil`, so that the `vim_item` got a `nil` value ?

CharlesChiuGit: hmmm, got it. seems like if `opts.before` is `false`, `vim_item` will return `nil`. Will send patch.

CharlesChiuGit: @kagaya85 plz check #652

PR #657, opened 2023-04-13
chore(mapping): cleanup

Jint-lzxy:
This commit made the following changes:
- Simplify binding commands (e.g., `map_cmd(":...<CR>") -> map_cr("...")`).
- **Remove** `<leader><leader>x` as this is only available on \*nix-like systems.
- **Remove** `<leader><leader>s` b/c some terminals may not refresh their UI in a _timely manner_ [^1]

[^1]: _The explanation may be a bit complicated, but I will try to keep it concise_:
Since this keymap starts with `<Cmd>`, although `nvim` **interprets** it as `:%s/\<(word-under-cursor)\>/(word-under-cursor)/gI` and **places** that on the command line, some terminals may _still_ refuse to refresh their UI _(usually, instructions starting with `<Cmd>` will **not** be displayed)_. Since we are, however, in command mode at this point, it would be like _typing in the dark_ - those terminals will only display the previous command after the user performs an operation, which may lead to further issues.
Sometimes `nvim` would detect this and throw _E5520: \<Cmd\> mapping must end with \<CR\>_


CharlesChiuGit (approved):
LGTM

PR #662, opened 2023-04-16
fix: error caused by treesitter entry_filter.

ayamir:
I got error under current config, and I don't know why a `entry_filter` is needed for treesitter source.

image

### DIFF
--- a/lua/modules/configs/completion/cmp.lua
+++ b/lua/modules/configs/completion/cmp.lua
@@ -151,11 +151,15 @@ return function()
 			{
 				name = "treesitter",
 				entry_filter = function(entry)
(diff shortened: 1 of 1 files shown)


CharlesChiuGit:
I have this error once before, but it was gone after I sync all plugins.

CharlesChiuGit:
The `entry_filter` is to filter out noisy suggestions from `cmp-treesitter`.

Jint-lzxy:
> I have this error once before, but it was gone after I sync all plugins.

Same here. This cannot be reproduced even in the same location after syncing the plugins.

ayamir:
ok, close it.

PR #669, opened 2023-04-19
fix: paint word boundary in python

fecet:
Current paint setup falsely recognize word boundary in python
image


CharlesChiuGit:
Sorry, my bad lol

PR #684, opened 2023-04-25
update transparent background setting in catppuccin.lua

AngelontheRoad:


CharlesChiuGit:
can u merge two pr in to one?
also plz use a better branch name, like `chore/settings`.

PR #685, opened 2023-04-25
fix(bufferline): use defaults for `get_element_icon`

Jint-lzxy:
This PR partially reverted commit ecaba008ae53e66113f0938a0605416de4d3ec83, fixing a regression.

Under current config _(using custom `get_element_icon`)_, all tabs will have the same icon _(since `vim.bo.filetype` can only hold one valid value)_


CharlesChiuGit:
omg, didn't notice that.
will test once i got home.

CharlesChiuGit (approved):
LGTM

PR #699, opened 2023-05-02
feat: make ui more tidy and unified

huge-pancake:
1. same appearance for cmp and wilder
2. clear appearance for bufferline working with nvim-tree and outline (lspsaga)
3. colored indent-blankline
4. similar highlights for indent lines in nvim-tree, indent-blankline and outline (lspsaga)

PR #707, opened 2023-05-04
feat(nvim-treesitter): add nvim-treesitter-context

Saafo:
close https://github.com/ayamir/nvimdots/discussions/693#discussioncomment-5802367

### DIFF
--- a/lua/modules/plugins/editor.lua
+++ b/lua/modules/plugins/editor.lua
@@ -82,6 +82,7 @@ editor["nvim-treesitter/nvim-treesitter"] = {
 	config = require("editor.treesitter"),
 	dependencies = {
 		{ "nvim-treesitter/nvim-treesitter-textobjects" },
+		{ "nvim-treesitter/nvim-treesitter-context" },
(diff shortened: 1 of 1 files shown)


CharlesChiuGit:
Don't we need to `require` it?

CharlesChiuGit:
Also, this might need to set the highlightings in #699 as well.

Saafo:
> Don't we need to `require` it? https://github.com/nvim-treesitter/nvim-treesitter-context#configuration

The default value is good enough to me, so we can just add to dependencies, not to require it.

> Also, this might need to set the highlightings in https://github.com/ayamir/nvimdots/pull/699 as well.

I've read this pr and I'm exciting to it! But I think it might out of this pr's scape since the highlighting pr is not merged yet.

CharlesChiuGit:
> I've read this pr and I'm exciting to it! But I think it might out of this pr's scape since the highlighting pr is not merged yet thinking.

make sense.

CharlesChiuGit (commented):
LGTM

PR #735, opened 2023-05-14
chore(keymap): lsp keymap should effective at `LspAttach` event

Groveer:
Lsp's keymap should not work when I'm editing a normal file。


ayamir:
https://github.com/ayamir/nvimdots/pull/699#issuecomment-1532371932

ayamir (approved):
Reasonable 👍

PR #738, opened 2023-05-15
feat!: use executables on `$PATH`

Groveer:


Jint-lzxy (changes requested):
I am not very supportive of making this change b/c we _ought to_ use the pre-compiled executables for `host_prog`s due to the `-m <mod>` flag (usually `-m pip <arguments>` is used for installing dependencies, and users who have installed many versions of Python(3) will encounter environment issues)

```console
-m mod : run library module as a script (terminates option list)
```

Groveer:
But using this change solves the problem of adaptation to different platforms. You cannot assume that the command exists in the specified directory, especially for the Windows platform.

Jint-lzxy:
IMO issuing errors early (executable not found) is way more intuitive (and debuggable) than getting `ModuleNotFoundError: No module named '<random-package-name>'`. b/c there're still many users who don't use `Python` (or are new to `nvim`) but still need `python_host_prog` support. FWIW simply changing the file path is easier than debugging an unfamiliar language.

Say if a user gets `ExecutableNotFoundError` first and then `ModuleNotFoundError` again, then things are very clear - that _specific_ Python distribution used doesn't have corresponding dependencies installed.

Groveer:
I don't have any doubts, but I still think that it would be better to use environment variables, one is more general, it can be cross-platform, and the other is that users should be responsible for their own environment variables, after all, setting environment variables is a basic operation for developers.

Jint-lzxy:
> I don't have any doubts, but I still think that it would be better to use environment variables, one is more general, it can be cross-platform, and the other is that users should be responsible for their own environment variables, after all, setting environment variables is a basic operation for developers.

Agree with you on this. We shall wait for other reviewers to check this PR, and if most of us are happy with this idea, we can also make more changes and remove several unnecessary conditionals 👍

ayamir:
For versatility, just use executable name is better because we provides cross-platform experience in fact. Approve these changes.

Jint-lzxy (approved):
LGTM

CharlesChiuGit:
side note: i hard coded the path just to get a faster startup time.

PR #736, opened 2023-05-15
chore(wsl): use Windows native way to yank and paste.

ayamir:
#731

PR #740, opened 2023-05-15
refactor: update wsl clipboard for windows with recomand

aloTu:
use the recommanded config to use clipboard in windows.
https://github.com/neovim/neovim/wiki/FAQ#how-to-use-the-windows-clipboard-from-wsl


CharlesChiuGit:
#736

PR #741, opened 2023-05-16
use clipboard for windows as nvim recommand

aloTu:
https://github.com/neovim/neovim/wiki/FAQ#how-to-use-the-windows-clipboard-from-wsl


CharlesChiuGit:
#736

PR #743, opened 2023-05-16
fix(global): wrong way of judging about the WSL environment

Groveer:
`vim.fn.has`method always return true，maybe should use `vim.fn.executable`.

### DIFF
--- a/lua/core/global.lua
+++ b/lua/core/global.lua
@@ -5,7 +5,7 @@ function global:load_variables()
 	self.is_mac = os_name == "Darwin"
 	self.is_linux = os_name == "Linux"
 	self.is_windows = os_name == "Windows_NT"
-	self.is_wsl = vim.fn.has("wsl") == 1
+	self.is_wsl = vim.fn.executable("wsl.exe") == 1
 	self.vim_path = vim.fn.stdpath("config")
 	local path_sep = self.is_windows and "\\" or "/"
 	local home = self.is_windows and os.getenv("USERPROFILE") or os.getenv("HOME")


Jint-lzxy (commented):
In fact this works on non-Windows systems. What is the issue you're encountering?

info

***

`:h feature-list`:
```console
List of supported pseudo-feature names:
<...>
wsl		WSL (Windows Subsystem for Linux) system.
```

Groveer:
> 
When I disable windows path in wsl environment, `has` method always return 1.
Like this in `wsl.conf`:
```ini
[interop]
enabled=false
appendWindowsPath=false
```
`has` method return 1, `executable` return 0, I think executable is correct.

CharlesChiuGit:
then what happens when u r outside of wsl and use nvim in native windows? doesn't executable return true?

Jint-lzxy (commented):
u mean you choose not to inherit `$PATH` from Windows _(parent system)_ inside wsl? But you are _still_ running ur Linux distribution inside wsl so `has("wsl")` should still return `1`, the purpose of this entry (`is_wsl`) is to verify whether the user is _using_ wsl, **not** whether he _has the executable_ or not (similar to checking `/proc/sys/fs/binfmt_misc/WSLInterop` , not `wsl.conf` ) so FWIW it is not appropriate to use `vim.fn.executable`.

Groveer:
> 

`has` return 0, and `executable` return 1, in native Windows. It seems that this change will cause problems in the native Windows .

Jint-lzxy:
> `has` return 0, and `executable` return 1, in native Windows. It seems that this change will cause problems in the native Windows .

Yep this is what we would like to check - the user is using Windows, but **not** using wsl.

Groveer:
> u mean you choose not to inherit `$PATH` from Windows _(parent system)_ inside wsl? But you are _still_ running ur Linux distribution inside wsl so `has("wsl")` should still return `1`, the purpose of this entry (`is_wsl`) is to verify whether the user is _using_ wsl, **not** whether he _has the executable_ or not (similar to checking `/proc/sys/fs/binfmt_misc/WSLInterop` , not `wsl.conf` ) so FWIW it is not appropriate to use `vim.fn.executable`.

I agree with you, but what to do in this situation, it will cause my clipboard to not work. Because the clipboard use Windows `$PATH`

Jint-lzxy:
So the root cause of ur issue is `nvim` failed to use the system's native clipboard located under Windows? This is b/c neovim cannot find the required executables:

https://github.com/ayamir/nvimdots/blob/986be6e18683f02bb3ed2c5183de82efbca34345/lua/core/init.lua#L102-L115

Please create soft links for `clip.exe` and `powershell.exe` on any of ur `$PATH` under Linux.

Groveer:
By the way, even though I turned off Windows `$Path` in the wsl, but I am using wl-clipboard, so far, I have not find any problems, and even the clipboard content can be transferred with the native windows.

Groveer:
> 

I don't think this is a correct solution, have you considered using neovim's default clipboard?

Groveer:
> So the root cause of ur issue is `nvim` failed to use the system's native clipboard located under Windows? This is b/c neovim cannot find the required executables:
> 
> https://github.com/ayamir/nvimdots/blob/986be6e18683f02bb3ed2c5183de82efbca34345/lua/core/init.lua#L102-L115
> 
> Please create soft links for `clip.exe` and `powershell.exe` on any of ur `$PATH` under Linux.

I don't think this is a correct solution, have you considered using neovim's default clipboard?

CharlesChiuGit:
I'm a bit confuse here. So u want to only use https://github.com/bugaevc/wl-clipboard in wsl, and not share clipboard content with `clip.exe`? And with current https://github.com/ayamir/nvimdots setup, u will have issues using `wl-clipboard` in nvim in wsl?

Jint-lzxy:
> By the way, even though I turned off Windows `$Path` in the wsl, but I am using wl-clipboard, so far, I have not find any problems, and even the clipboard content can be transferred with the native windows.

Yep this _indeed_ works fine with recently-released wsl, but we aim to provide support for any version of wsl. In fact, creating soft links for executables on Windows when wsl does not inherit its `$PATH` is used in many places to provide easy-fixes for environment issues.

Jint-lzxy:
> have you considered using neovim's default clipboard

Yep but users could encounter weird behaviors when they're using older builds of wsl.

Jint-lzxy:
> have you considered using neovim's default clipboard

Ys, but users could encounter weird behaviors when they're using older builds of wsl 😄

Groveer:
> 

yes, nvim will tell me cannot find `clip.exe` and `powershell.exe`, like this:
图片

CharlesChiuGit:
> I'm a bit confuse here. So u want to only use [bugaevc/wl-clipboard](https://github.com/bugaevc/wl-clipboard) in wsl, and not share clipboard content with `clip.exe`? But with current [ayamir/nvimdots](https://github.com/ayamir/nvimdots) setup, u will have issues using `wl-clipboard` in nvim under wsl?
> 
> If that's the case, I would suggest u to fork this repo and use your custom clipboard settings.

To me, it seems more like a custom clipboard issue, rather than how we find if we're in wsl or not.

Groveer:
> I'm a bit confuse here. So u want to only use https://github.com/bugaevc/wl-clipboard in wsl, and not share clipboard content with `clip.exe`? But with current https://github.com/ayamir/nvimdots setup, u will have issues using `wl-clipboard` in nvim under wsl?
> 
> If that's the case, I would suggest u to fork this repo and use your custom clipboard settings.

yes, nvim will tell me not find clip.exe, like this:
图片

Groveer:
> I'm a bit confuse here. So u want to only use https://github.com/bugaevc/wl-clipboard in wsl, and not share clipboard content with `clip.exe`? But with current https://github.com/ayamir/nvimdots setup, u will have issues using `wl-clipboard` in nvim under wsl?
> 
> If that's the case, I would suggest u to fork this repo and use your custom clipboard settings.

it seems that there is no other choice.

CharlesChiuGit:
> yes, nvim will tell me not find clip.exe, like this:
> 图片

Reasonable, since u disable the windows `$PATH` in your wsl.

Jint-lzxy:
Then how to address this issue is totally up to you - u may change clipboard configs or create soft links.

Groveer:
Ok, you can close this PR.

Issue #742, opened 2023-05-16
Adapt to lazy load session_lens of auto-session


Saafo:
### Feature description

Currently, [auto-session](https://github.com/rmagatti/auto-session) start with configuring session_lens, so that if we start with auto-session, it will load telescope immediately and will cost about 40 ms at startup.

Now, it fixed this issue by https://github.com/rmagatti/auto-session/pull/231, and we can configure this for auto-session:

```
  session_lens = {
    load_on_setup = false,
  },
```

to avoid loading telescope at start, which will only cost about 1ms even if we load auto-session at startup.

This issue is suggesting adding the configuration above to the auto-session config. And if possible, I can create a pr for it.

Besides, I wonder whether we should call `require("auto-session").setup_session_lens()` at some point before using session-lens

ayamir:
We have lazy loaded the `auto-session` itself with `SessionSave`, `SessionRestore` and `SessionDelete` cmds.
https://github.com/ayamir/nvimdots/blob/42e816b9d465b69110de639b1afd94b3ce1fa5c8/lua/modules/plugins/editor.lua#L10
It does't affect the startuptime.

Saafo:
> We have lazy loaded the auto-session itself with SessionSave, SessionRestore and SessionDelete cmds.

Oh I mean, if someone (like me) enable the `auto-session` by `lazy=false`, then the configuration helps.

ayamir:
```lua
return function()
	require("auto-session").setup({
		log_level = "info",
		auto_session_enable_last_session = true,
		auto_session_root_dir = vim.fn.stdpath("data") .. "/sessions/",
		auto_session_enabled = true,
		auto_save_enabled = true,
		auto_restore_enabled = true,
		auto_session_suppress_dirs = nil,
		session_lens = {
			-- If load_on_setup is set to false, one needs to eventually call `require("auto-session").setup_session_lens()` if they want to use session-lens.
			load_on_setup = false,
			theme_conf = { border = true },
			previewer = false,
		},
	})

	vim.api.nvim_create_user_command("SessionLenShow", function()
		require("auto-session").setup_session_lens()
		require("auto-session.session-lens").search_session()
	end, { nargs = 0 })
end

```

Saafo:
LGTM!

PR #745, opened 2023-05-16
feat: load `SessionLens` manually via user command

ayamir:
https://github.com/ayamir/nvimdots/issues/742

### DIFF
--- a/lua/modules/configs/editor/auto-session.lua
+++ b/lua/modules/configs/editor/auto-session.lua
@@ -7,5 +7,16 @@ return function()
 		auto_save_enabled = true,
 		auto_restore_enabled = true,
 		auto_session_suppress_dirs = nil,
+		session_lens = {
+			-- If load_on_setup is set to false, one needs to eventually call `require("auto-session").setup_session_lens()` if they want to use session-lens.
+			load_on_setup = false,
+			theme_conf = { border = true },
+			previewer = false,
+		},
 	})
+
+	vim.api.nvim_create_user_command("SessionLenShow", function()
+		require("auto-session").setup_session_lens()
+		require("auto-session.session-lens").search_session()
+	end, { nargs = 0 })
 end
--- a/lua/modules/configs/editor/auto-session.lua
+++ b/lua/modules/configs/editor/auto-session.lua
@@ -8,14 +8,14 @@ return function()
 		auto_restore_enabled = true,
 		auto_session_suppress_dirs = nil,
 		session_lens = {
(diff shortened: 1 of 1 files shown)


ayamir:
A more reasonable way is judge if setuped when execute `SessionLenShow` but I have not idea to get the condition about it. Maybe better solution is provide similar command in upstream itself.

Jint-lzxy (dismissed):
I am hesitant to merge this PR into `main` b/c we have already set `auto-session` to load-on-demand. Perhaps @Saafo can add this to his/her custom config 👍

> A more reasonable way is judge if setuped

We have [`package.loaded`](https://www.lua.org/manual/5.1/manual.html#pdf-package.loaded) :)

Saafo:
> Perhaps @Saafo can add this to his/her custom config 👍

Both are fine to me!

ayamir:
> We have [package.loaded](https://www.lua.org/manual/5.1/manual.html#pdf-package.loaded) :)

Seems I can't set it like below.
```lua
	vim.api.nvim_create_user_command("SessionLenShow", function()
		if not package.loaded("telescope") then
			require("auto-session").setup_session_lens()
		end
		require("auto-session.session-lens").search_session()
	end, { nargs = 0 })
```

Jint-lzxy:
> Seems I can't set it like below because the setup status of `session_lens` have no direct relation of whether telescope loaded or not.

We could check `auto-session.session-lens` instead ;)

ayamir:
I tend to merge it into main.

lua/modules/configs/editor/auto-session.lua
```diff
@@ -7,5 +7,18 @@ return function()
 		auto_save_enabled = true,
 		auto_restore_enabled = true,
 		auto_session_suppress_dirs = nil,
+		session_lens = {
+			-- If load_on_setup is set to false, one needs to eventually call `require("auto-session").setup_session_lens()` if they want to use session-lens.
```
Jint-lzxy, line 11:
```suggestion
			-- If `load_on_setup` is set to false, please use `SessionLensToggle` to manually load this plugin.
```

lua/modules/configs/editor/auto-session.lua
```diff
@@ -7,5 +7,18 @@ return function()
...
+			load_on_setup = false,
+			theme_conf = { border = true },
+			previewer = false,
+		},
 	})
+
+	vim.api.nvim_create_user_command("SessionLenShow", function()
```
Jint-lzxy, line 18:
```suggestion
	vim.api.nvim_create_user_command("SessionLensToggle", function()
```

Jint-lzxy (commented):
Adjusted phrasing 👍

Issue #750, opened 2023-05-18
Add `FormatToggle` keymap


Groveer:
### Feature description

I want make format on save disable by default, but I want enable it some time. Maybe we should add a keymap for `FormatToggle`. look #606 and #347

CharlesChiuGit:
See, the difficulty of making a keymap is defining a proper and memorable one. U can propose one first.

PR #751, opened 2023-05-18
feat: add `FormatToggle` keymap (#750)

Groveer:


CharlesChiuGit:
Functionally this should works, but I'm not sure how to feel about this.

Groveer:
I'm a C/C++ developer, I do't need `format on save` by default, it will modify my source file all. so, I set format on save to false, but sometimes when I edit html or md file, I need it auto format, so, I need a keymap to toggle it.

CharlesChiuGit:
have u consider just put your project path in the disable formatting settings? check lua\settings.lua

CharlesChiuGit:
image

CharlesChiuGit:
i don't think everything needs to be send to upstreams. in fact, most of the maintainers of this repo keeps their own forks/repos to cater their needs. 

u can fork this repo, create another branch, and cherry-pock the commits to merge into your own branch.

CharlesChiuGit:
i'm leaning towards close this pr.

Groveer:
> 

This settings means that every project I need to disable auto format must add it‘s directory here, which is far less convenient than keymap.

Groveer:
> 

I agree with you, but I think this feature is more user-friendly.

CharlesChiuGit:
> This settings means that every project I need to disable auto format must add it‘s directory here, which is far less convenient than keymap.

Personally I just put all my projects in a folder called `Workspace`. If I were u, I could just add `~/Workspace` to the list.

ayamir (changes requested):
Set a keymap for `FormatToggle` is fine, but why need to modify `formatting.lua`?

Groveer:
Because there is bug, when I toggle to disable, nvim will not send a notification(toggle to enable wiil do this). modify `formatting.lua` will fix it.

ayamir:
> Because this is a bug, when I toggle to disable, nvim will not send a notification(toggle to enable wiil do this). modify `formatting.lua` will fix it.

Just add keymap works for me, never bug encountered.

Jint-lzxy:
> Because this is a bug, when I toggle to disable, nvim will not send a notification(toggle to enable wiil do this). modify `formatting.lua` will fix it.

In fact this is intentional. If one disables a config entry _(such as `format_on_save`)_ in `settings.lua`, we would assume he has **blocked** that feature _(just like supplying `-U<feature_macro_name>` during compilation)_. We do **not** expect users to use that feature in any way in the future, so there is no _guarantee_ for any related behaviors.

Jint-lzxy:
I'm also leaning towards closing this pr -

> I'm a C/C++ developer, I do't need `format on save` by default, it will modify my source file all. so, I set format on save to false, but sometimes when I edit html or md file, I need it auto format, so, I need a keymap to toggle it.

u might be interested in [Clang-Format Style Options](https://clang.llvm.org/docs/ClangFormatStyleOptions.html), and if you really want to disable code formatting for `c/cpp`, you can remove the corresponding formatter(s):
https://github.com/ayamir/nvimdots/blob/0f79d017ddbfbbe651cd96364418f4c8b32cca20/lua/modules/configs/completion/null-ls.lua#L12-L15

FWIW adding too many keymaps that are "seemingly" rare can make the config too cumbersome and further limit customizability. For features that have _strong practicality_ for _specific_ groups of users, they should be left to the users to implement those features _themselves_, which is also one of the reasons why we use `neovim` 😄

Groveer:
> > Because this is a bug, when I toggle to disable, nvim will not send a notification(toggle to enable wiil do this). modify `formatting.lua` will fix it.
> 
> In fact this is intentional. If one disables a config entry _(such as `format_on_save`)_ in `settings.lua`, we would assume he has **blocked** that feature _(just like supplying `-U<feature_macro_name>` during compilation)_. We do **not** expect users to use that feature in any way in the future, so there is no _guarantee_ for any related behaviors.

Thank you for explaining, it's up to you to merge the PR or not.

ayamir (approved):

Issue #612, opened 2023-04-03
Lost file history


fecet:
### Version confirmation

- [X] Confirm

### Following prerequisites

- [X] Confirm

### Neovim version

NVIM v0.8.3

### Operating system/version

Arch Linux

### Terminal name/version

Alacritty

### Branch info

main (Default/Latest)

### Fetch Preferences

HTTPS (use_ssh = false)

### How to reproduce the issue

Open and close some files

### Expected behavior

`Telescope oldfiles` list all files just opened

### Actual behavior

Only show the last one file

Jint-lzxy:
Cannot reproduce.

Can u provide a _detailed_ repro? You may use a pseudonym instead of a real filename (e.g., use `A.lua` instead of `init.lua`)
***
Example:
1. Use Telescope to open `./A.lua`
2. Then use nvim-tree to open `./A/B.txt`
3. Close `./A/B.txt`
4. `Telescope oldfiles` shows only one of the files (i.e., `./A.lua`)

fecet:
I just find it's not relevant to `telescope` since
```
:oldfiles
```
only list
```
1: /home/rok/Dashboard-1
2: /home/rok/[Wilder Float 0]
3: /home/rok/[Wilder Float 1]
```

Jint-lzxy:
So you mean `Telescope oldfiles` didn't list the files opened in previous `nvim` instances? Then this is _exactly_ how it works - providing file access within **the current process**, which is often used to reopen _previously opened_ files by [scanning all available buffers](https://github.com/nvim-telescope/telescope.nvim/blob/942fe5faef47b21241e970551eba407bc10d9547/lua/telescope/builtin/__internal.lua#L500-L547).

But FYI, the implementation of `:oldfiles` is actually somewhat different from that of `Telescope oldfiles`:
```console
						*:o* *:ol* *:oldfiles*
:o[ldfiles]		List the files that have marks stored in the ShaDa
			file.  This list is read on startup and only changes
			afterwards with `:rshada!`.  Also see |v:oldfiles|.
			The number can be used with |c_#<|.
			The output can be filtered with |:filter|, e.g.:
				filter /\.vim/ oldfiles
			The filtering happens on the file name.
```

[...]

PR #769, opened 2023-05-26
fix: merge shada after VimLeave, #612.

ayamir:


Jint-lzxy (changes requested):
What I mean in #612 is we should modify `'shada'` (the option). In fact, the current issue is previous ShaDa file(s) won't get merged with the ShaDa in memory since we supplied `!` (forced):

```console
						    *:wsh* *:wshada* *E137*
:wsh[ada][!] [file]	Write to ShaDa file [file] (default: see above).
			The information in the file is first read in to make
			a merge between old and new info.  When [!] is used,
			the old information is not read first, only the
			internal info is written (also disables safety checks
			described in |shada-error-handling|).  If 'shada' is
			empty, marks for up to 100 files will be written.
			When you get error "E929: All .tmp.X files exist,
			cannot write ShaDa file!" check that no old temp files
			were left behind (e.g.
			~/.local/state/nvim/shada/main.shada.tmp*).

			Note: Executing :wshada will reset all |'quote| marks.
```

Jint-lzxy (commented):
@ayamir @fecet Is `Telescope oldfiles` working properly now?

lua/core/event.lua
```diff
@@ -114,11 +114,11 @@ function autocmd.load_autocmds()
 				"*",
 				[[if &cursorline && &filetype !~# '^\(dashboard\|clap_\)' && ! &pvw | setlocal nocursorline | endif]],
 			},
-			-- Force write shada on leaving nvim
+			-- Attempt to write shada when leaving nvim
 			{
 				"VimLeave",
 				"*",
-				[[if has('nvim') | wshada! | else | wviminfo! | endif]],
+				[[if has('nvim') | wshada | else | wviminfo! | endif]],
```
ayamir, line 121: Just set this line `"wsl"` fine.

fecet:
> @ayamir @fecet Is `Telescope oldfiles` working properly now?

I have set this to "wshada" and everything works fine since that time.

lua/core/event.lua
Jint-lzxy: This is to prevent strange issues with some self-built `nvim`s :( Since `wviminfo!` is always available, but `wshada` may not.

ayamir: OK, LGTM.

Jint-lzxy (approved):

Issue #767, opened 2023-05-26
Neodim not work for pylsp


fecet:
### Version confirmation

- [X] Confirm

### Following prerequisites

- [X] Confirm

### Neovim version

NVIM v0.9.0

### Operating system/version

Arch

### Terminal name/version

alacritty

### Branch info

main (Default/Latest)

### Fetch Preferences

HTTPS (use_ssh = false)

### Affected language servers

pylsp, pyright

### How to reproduce the issue

open any python file, https://github.com/zbirenbaum/neodim/issues/21#issuecomment-1557531973
```python
from typing import Union, List, Any


def a(b, c) -> Any:
    return b + 1
```
And the following is how it look likes in neovim:
![image](https://github.com/zbirenbaum/neodim/assets/41792945/6fdca40a-47e4-480a-8dd0-b1440964b13d)

### Actual behavior

Neodim not dim unused variable

### Support info

 Language client log: /home/rok/.local/state/nvim/lsp.log
 Detected filetype:   python
 
 2 client(s) attached to this buffer: 
 
 Client: pylsp (id: 1, bufnr: [1])
 	filetypes:       python
 	autostart:       true
 	root directory:  /home/rok/Nutstore Files/codes/tlem
 	cmd:             /home/rok/.local/share/nvim/mason/bin/pylsp
 
 Client: null-ls (id: 2, bufnr: [1])
 	filetypes:       cpp, c, vue, typescript, javascript, typescriptreact, javascriptreact, yaml, css, html, sh, markdown, scss, vim, python, lua, luau, rust
 	autostart:       false
 	root directory:  /home/rok/Nutstore Files/codes/tlem
 	cmd:             <function>
 
 Configured servers list: html, jsonls, lua_ls, clangd, ruby_ls, pylsp, bashls

ayamir:
Can confirm, Maybe `neodim` doesn't support `pylsp`.

PR #770, opened 2023-05-26
feat: set default config for pylsp.

ayamir:
I set the default config for `pylsp` and solved the problem in https://github.com/ayamir/nvimdots/issues/767 by enable `ruff` to do lint.

Main config:
1. Use [python-lsp-ruff](https://github.com/python-lsp/python-lsp-ruff) to provide lint function due to its high performance.
2. Use [pylsp-rope](https://github.com/python-rope/pylsp-rope) to provide code refactor function in code action's way as mentioned in https://github.com/ayamir/nvimdots/discussions/708#discussioncomment-5890640.
3. Use [python-lsp-black](https://github.com/python-lsp/python-lsp-black) to provide format function.
4. Remove black config in `null-ls`.
5. Users no need to install or update these plugins manually after `pylsp` installed or updated.

The result: 
image

### DIFF
--- a/lua/modules/configs/completion/servers/pylsp.lua
+++ b/lua/modules/configs/completion/servers/pylsp.lua
@@ -6,20 +6,33 @@ return {
 				-- lint related
 				ruff = {
 					enabled = true,
+					select = {
+						-- enable pycodestyle
+						"E",
+						-- enable pyflakes
+						"F",
+					},
+					ignore = {
+						-- ignore E501 (line too long)
+						"E501",
+						-- ignore F401 (imported but unused)
+						"F401",
+					},
 					extendSelect = { "I" },
 				},
 				-- refactor related
 				rope = { enabled = true },
 				-- format related
 				black = { enabled = true },
-				pyls_isort = { enabled = true },
 
 				-- disabled tools
 				-- lint related
 				flake8 = { enabled = false },
 				pyflakes = { enabled = false },
 				pycodestyle = { enabled = false },
+				mccabe = { enabled = false },
 				-- format related
+				pyls_isort = { enabled = false },
 				autopep8 = { enabled = false },
 				yapf = { enabled = false },
 			},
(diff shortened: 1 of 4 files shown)


ayamir:
@fecet @CharlesChiuGit

fecet:
I just started learning about Python's linter and formatter today, so I may not be able to provide any assistance beyond testing. 
What could be the reasons for neodim not supporting it? Ruff support both format and lint. How about using ruff provided by ruff-lsp or null-ls? I've seen https://github.com/CharlesChiuGit/nvimdots.lua/blob/df6576052e3d72f8a76e59457c0101203c634d42/lua/core/settings.lua#L109
configure it that way, if I understand correctly.

ayamir:
I tend to use `ruff` as plugin of `pylsp` because the dependency is linear.

fecet:
I have tried this PR and everything works fine! Except I notice two things:

- python-lsp-ruff lint seems too strict, sources like "import but unused" or "line too long" would be considered as "Error".
- ruff's code action "Ruff organize import" will randomly move character:
image

Of course all of them are upstream issues.

ayamir:
1. I added `F401` and `E501` to ignore.
2. Disable isort because `ruff` can do the same thing.

fecet:
I think it's better not to include any ignore as default, if anyone found this was annoyed, they could add that in ther `pyproject.toml` or set  `diagnostics_virtual_text` as false (I did this). As there is hope for this issue to be resolved quickly per https://github.com/python-lsp/python-lsp-ruff/issues/36

ayamir:
OK，we can wait for this.

CharlesChiuGit:
Agree with @fecet , we can just comment out the `ignore` table.

lua/modules/configs/completion/servers/pylsp.lua
```diff
@@ -0,0 +1,41 @@
...
+					enabled = true,
+					select = {
+						-- enable pycodestyle
+						"E",
+						-- enable pyflakes
+						"F",
+					},
+					ignore = {
+						-- ignore E501 (line too long)
+						"E501",
```
CharlesChiuGit, line 17: I think it's better to comment both `E501` and `F401`.

CharlesChiuGit (changes requested):

CharlesChiuGit (approved):

PR #807, opened 2023-06-19
fix(ui): telescope ui not adapt transparent background

Groveer:


CharlesChiuGit (approved):
LGTM, this also not interference with current buffer. Looks pretty great.

lua/modules/configs/ui/catppuccin.lua
```diff
@@ -140,15 +140,41 @@ return function()
...
+					TelescopePromptTitle = {
+						fg = transparent_background and cp.red or cp.base,
+						bg = transparent_background and cp.none or cp.red,
+					},
+					TelescopeResultsTitle = {
+						fg = cp.mantle,
+						bg = transparent_background and cp.none or cp.mantle,
+					},
+					TelescopeSelection = {
+						fg = cp.text,
```
ayamir, line 175: Set it to `cp.green` to keep consistency with selection of cmp.

ayamir (changes requested):

lua/modules/configs/ui/catppuccin.lua
ayamir: image

ayamir: Of course, it will also affect the fg color without transparent background like this. It's acceptable IMO.
image

Groveer: 。Good idea! already changed.

PR #808, opened 2023-06-19
feat(scrollview): enable signs support

Jint-lzxy:
[nvim-scrollview#signs](https://github.com/dstein64/nvim-scrollview#signs) and https://github.com/ayamir/nvimdots/discussions/494#discussioncomment-6076909


Jint-lzxy:
But the signs on RHS (or beside the scrollbar) uses _relative positions_ (to the document), FWIW _the signs showed on the left_ are still required for precise positioning.

Issue #347, opened 2022-12-05
how to disable FormatOnSave permanently?


goodpaperman:
### Neovim version

NVIM v0.8.1 Build type: RelWithDebInfo LuaJIT 2.1.0-beta3

### Operating system/version

CentOS Linux release 7.9.2009 (Core)

### Terminal name/version

mac term

### $TERM environment variable

xterm-256color

### Branch info

main (Default/Latest)

### Fetch Preferences

SSH (use_ssh = true)

### Affected language servers

clang-format

### How to reproduce the issue

:w save file，code auto formatted

### Actual behavior

code formatted

### Expected behavior

not format

### Support info

:LspInfo
![图片](https://user-images.githubusercontent.com/8522640/205540261-51dee403-f5b7-4c7e-88ac-a6e74edb311c.png)

:Mason
![图片](https://user-images.githubusercontent.com/8522640/205540312-d5be2812-a875-476b-893e-2213ff3102e9.png)

### Logs

```console
no logs find in given path
```

### Additional information

using FormatToggle can disable this, but sometime forget to set this, can I disable this feature in config file?

ayamir:
here: https://github.com/ayamir/nvimdots/blob/63113e2bb2031c335ce7ab70799dc24502fd27d3/lua/core/settings.lua#L5

goodpaperman:
it works:

```
local settings = {}
local home = os.getenv("HOME")

settings["use_ssh"] = true
settings["format_disabled_dirs"] = {
	home .. "/format_disabled_dir_under_home",
    home .. "/code/gevent",
    "/home/ext/project/baidu/netdisk/p2p-sdk-mobile/project/src"
}

return settings
```

can I specify things like 'code/*' to do regular express match?

ayamir:
currently can't

goodpaperman:
can I disable clang-format on start?

ayamir:
comment this line:
https://github.com/ayamir/nvimdots/blob/63113e2bb2031c335ce7ab70799dc24502fd27d3/lua/modules/completion/lsp.lua#L273

goodpaperman:
lsp complains about this:

![图片](https://user-images.githubusercontent.com/8522640/205565392-1d8a1019-4841-4c71-8ccf-3b09d48c0a33.png)

ayamir:
So what do you want to do?

goodpaperman:
[...]

PR #813, opened 2023-06-21
feat(formatting): format disabled dirs may contain regexps

Jint-lzxy:
This PR corrected `format_disabled_dirs`'s behavior - we should obtain full path of the _current_ file instead of the internal directory returned by `getcwd()`, as sometimes they may be inconsistent (e.g., after `:[lt]cd`); For `M.format` implemented regexp support (requested [here](https://github.com/ayamir/nvimdots/issues/347#issuecomment-1336687331)) and automatic normalization of given paths - more testing on different platforms might be necessary.

https://github.com/ayamir/nvimdots/assets/50296129/22381e32-6b65-4329-8f86-1d2462cf9d76

> Commands used:
> ```console
> :w
> :tcd ~/Desktop
> :w
> ```

Possible issues that may arise are:
- Escape of special characters may become invalid (`\\` will be automatically converted to `/`, for example, `[[hello\+]]` will become `[[hello/+]]`)
- Directories provided by the user may be ill-formed after normalization.

### DIFF
--- a/lua/core/settings.lua
+++ b/lua/core/settings.lua
@@ -25,7 +25,7 @@ settings["diagnostics_virtual_text"] = true
 settings["diagnostics_level"] = "Hint"
 
 -- Set the format disabled directories here, files under these dirs won't be formatted on save.
---- NOTE: String(s) may contain regular expressions (grammar: vim). |regexp|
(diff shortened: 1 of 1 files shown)


ayamir (approved):
LGTM. 👍

CharlesChiuGit (approved):
LGTM

Issue #698, opened 2023-05-02
Tidy and unified UI


volpan:
Straightly go to [this comment](https://github.com/ayamir/nvimdots/issues/698#issuecomment-1585716290).

### Feature description

The highlights for cmp, nvim-tree, wilder and indent-blankline now is untidy and always catch my focus when I should focus on my code.

### Additional information

I had already modified the highlights for them.

1. same appearance for cmp and wilder
2. clear appearance for bufferline and nvim-tree
3. colored indent-blankline
4. similar highlights for indent lines in nvim-tree, indent-blankline and outline of lspsaga

![image](https://user-images.githubusercontent.com/73375859/235570767-151f32b2-bef2-415d-a5f8-04b794ecb230.png)

![image](https://user-images.githubusercontent.com/73375859/235570884-5c7f5212-d526-4684-9242-70a953f41795.png)

So I'm here to get to know if I could give a PR.

huge-pancake:
update: highlight for lspsaga's outline [lspsaga#1050](https://github.com/nvimdev/lspsaga.nvim/issues/1050#issue-1691710152)

![image](https://user-images.githubusercontent.com/73375859/235578477-68bbfb06-0916-48fa-8fa9-70b83b869234.png)

CharlesChiuGit:
@Jint-lzxy 's #676 PR should alleviate some of your issues. It's most from upstream, so it needs more time to finish it.

huge-pancake:
You mean some of the problems are from `catppuccin/nvim` right? So I will wait for them :P

They other changes are not problems, they are just for customization (2, 3 and 4)

CharlesChiuGit:
ys, of course u can send pr about 2, 3, 4 first!

Jint-lzxy:
@huge-pancake Many thanks for your contribution 👍  But due to the complexity of #676 (most highlight groups and _actual hex codes_ for those _antonomasias_ \[e.g., `C.base -> #1E1E2E`\] will undergo significant changes), hence to avoid repetitive work, this PR will be merged after #676 is completed. I will also refer to your PR while reimplementing those hl groups.

Here is a sneak peak of Telescope's new UI _(based on your PR)_:

<img width="1792" alt="telescope-new-ui" src="https://user-images.githubusercontent.com/50296129/235711540-0e076386-8e94-4856-adc4-d6b37a6ecafe.png">

huge-pancake:
Got it, hope #676/upstream update will come out soon.

[...]

PR #831, opened 2023-06-30
feat(utils): auto-refresh builtin palette on `ColorScheme`

Jint-lzxy:
This PR implemented dynamic palette support for https://github.com/ayamir/nvimdots/issues/698#issuecomment-1614699356.

cc @volpan

### DIFF
--- a/lua/modules/utils/init.lua
+++ b/lua/modules/utils/init.lua
@@ -29,12 +29,30 @@ local M = {}
 ---@field crust string
 ---@field none "NONE"
 
----@type palette
+---@type nil|palette
 local palette = nil
+
+-- Indicates if the autocmd for refreshing the palette has already been registered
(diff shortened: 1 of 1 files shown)


volpan:
Then how can I use it in configs such as `lualine`?

Jint-lzxy:
Just use it as usual. The only difference is each call to `get_palette` would now return a palette configured for the active colorscheme - the same as `require("catppuccin.palettes").get_palette()`.

volpan:
Actually, if I use this in the autocmd `ColorScheme` (the same autocmd as util), it has a delay which means this can be applied in the next color-scheme change.

image

volpan:
LGTM

volpan (approved):

ayamir (approved):

PR #835, opened 2023-07-02
fix(options): correct scroll behaviors of (h)splits

Jint-lzxy:
Just found out that we removed `stabilize.nvim` but forgot to set the corresponding option :)


CharlesChiuGit (approved):
LGTM

ayamir (approved):

PR #837, opened 2023-07-03
chore(assets): put `dots.tutor` under the rtp

Jint-lzxy:
This commit puts `dots.tutor` under the rtp `tutor/` - this way, users can enjoy the ingenious manual through `:Tutor dots` without specifying any filepath 😄


CharlesChiuGit (approved):
I prefer this too.

ayamir (approved):

PR #838, opened 2023-07-03
perf(lualine): more efficient `has_git` checks

Jint-lzxy:
This works b/c `finddir()` searches bidirectionally _(downward searches are recursive and may search through many directories)_.

This commit instructs `fs.find` to only traverse upward through parent directories and terminate as long as a match is found.


ayamir (approved):
LGTM 👍

PR #840, opened 2023-07-05
workaround(lsp-signature)!: fix input delay of nvim-cmp

Jint-lzxy:
* `on_attach` [is deprecated](https://github.com/ray-x/lsp_signature.nvim/blob/17ff7a405fea8376b015b8ea7910d2e59958bf68/doc/lsp_signature.txt#L148-L149);
* Now `lsp_signature` won't adjust its position relative to the completion popup to avoid problematic functions (cmp.visible) being called.

### DIFF
--- a/lua/modules/configs/completion/lsp-signature.lua
+++ b/lua/modules/configs/completion/lsp-signature.lua
@@ -1,0 +1,15 @@
+return function()
+	require("lsp_signature").setup({
+		bind = true,
+		-- TODO: Remove the following line when nvim-cmp#1613 gets resolved
+		check_completion_visible = false,
+		floating_window = true,
+		floating_window_above_cur_line = true,
+		hi_parameter = "Search",
+		hint_enable = true,
+		transparency = nil, -- disabled by default, allow floating win transparent value 1~100
+		wrap = true,
+		zindex = 45, -- avoid overlap with nvim.cmp
+		handler_opts = { border = "single" },
+	})
+end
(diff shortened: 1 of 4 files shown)


ayamir (approved):

CharlesChiuGit (approved):
LGTM

YuCao16:
Does this fix work for you? Because I'm still experiencing input delay on my end. If I disable `lsp_signature`, input delay disappears.

Jint-lzxy:
@YuCao16 This PR only reduces calls to `cmp.visible()`, but as a matter of fact, u can completely eliminate related calls (since we disabled `check_completion_visible`) via this patch:
```diff
diff --git a/lua/lsp_signature/helper.lua b/lua/lsp_signature/helper.lua
index 8d09a22..c3e32d6 100644
--- a/lua/lsp_signature/helper.lua
+++ b/lua/lsp_signature/helper.lua
@@ -697,7 +697,7 @@ end
 helper.completion_visible = function()
   local hascmp, cmp = pcall(require, 'cmp')
   if hascmp then
-    return cmp.visible()
+    return true
   end

   return fn.pumvisible() ~= 0
```
The issue upstream is not _every_ call to `cmp.visible()` gets checked.

Jint-lzxy:
Will #890 fix that?

YuCao16:
> Will #890 fix that?

Yes, it works, thanks!

PR #845, opened 2023-07-07
migrate: install `rustfmt` from `rustup` (#842)

CharlesChiuGit:
Should we include `rustup` installation in our installation scripts?
Besides `rustfmt`, `sniprun` also needs `cargo` to compile, so I think it might be acceptable to add it.
cc @Jint-lzxy @ClSlaid @volpan

### DIFF
--- a/lua/core/settings.lua
+++ b/lua/core/settings.lua
@@ -99,7 +99,6 @@ settings["lsp_deps"] = {
 settings["null_ls_deps"] = {
 	"clang_format",
 	"prettier",
-	"rustfmt",
 	"shfmt",
 	"stylua",
 	"vint",


Jint-lzxy:
Sure! But how can we ensure proper installation (of `rustup`) on non-Windows platforms?

ayamir:
> Sure! But how can we ensure proper installation (of `rustup`) on non-Windows platforms?

IMO the most direct way is add `rustup` to [wiki](https://github.com/ayamir/nvimdots/wiki/Prerequisites).

CharlesChiuGit:
> Sure! But how can we ensure proper installation (of `rustup`) on non-Windows platforms?

I think we can just check if `cargo` and `rustc` is installed?

CharlesChiuGit:
> IMO the most direct way is adding `rustup` to [wiki/Prerequisites](https://github.com/ayamir/nvimdots/wiki/Prerequisites).
> 
> EDIT: for rust users

OK, I had added it in the prerequisites.

ayamir:
`cargo`, `rustc` and `rustfmt` are added by default when installing toolchain:
image

PR #852, opened 2023-07-10
fix(null-ls): specify `rustfmt` manually

Jint-lzxy:
Haven't introduced `rustup` into the install script(s) yet cause I can't figure out how to automate this process 😢

### DIFF
--- a/lua/modules/configs/completion/null-ls.lua
+++ b/lua/modules/configs/completion/null-ls.lua
@@ -25,6 +25,7 @@ return function()
 				"markdown",
 			},
 		}),
+		btns.formatting.rustfmt,
 	}
 	null_ls.setup({
 		border = "rounded",
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -397,8 +397,11 @@ Please make sure you have nvim v$REQUIRED_NVIM_VERSION_LEGACY installed at the v
 	}
 
 	info -Msg "Spawning Neovim and fetching plugins... (You'll be redirected shortly)"
-	info -Msg 'To make sqlite work with lua, manually grab the dlls from "https://www.sqlite.org/download.html" and'
-	info_ext -Msg 'replace vim.g.sqlite_clib_path with your path at the bottom of `lua/core/options.lua`'
+	info -Msg 'To make sqlite work with lua, manually grab the dlls from "https://www.sqlite.org/download.html" and replace'
+	info_ext -Msg 'vim.g.sqlite_clib_path with your path at the bottom of `lua/core/options.lua`'
+	info -Msg 'Also, please make sure you have a Rust Toolchain installed via `rustup`! Otherwise, unexpected things may'
+	info_ext -Msg 'happen. See: https://www.rust-lang.org/tools/install         ¯¯¯¯¯¯¯¯¯¯¯¯¯ '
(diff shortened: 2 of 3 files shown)


CharlesChiuGit:
https://github.com/CharlesChiuGit/dotfiles/blob/main/install_scripts/langs/rust.sh

maybe this would help

Jint-lzxy:
The problem is we probably need to avoid installing multiple versions of Rust in parallel - so things become much more complex here, as Rust is distributed in multiple ways:
- **Self-hosted distribution** (some package managers e.g., [brew](https://formulae.brew.sh/formula/rust#default)): the package manager manages version control, `rustfmt` must be installed via `cargo`, but this has been deprecated; Besides, Rust is usually NOT installed on a standard path (aka `~/.cargo`).
- **The "legacy" official installer**: Similar to the official installer, except that `rustup` won't be installed (by default); `rustfmt` must be installed via `cargo`, or via `rustup` but `rustup` itself must be installed first.
- **The official installer, or the `rustup[-init]` package provided by many package mgrs**: The recommended approach.

So if we decide to automate this process, the following should be considered:
- **How to cleanup legacy installation(s)?** This seems to be far beyond the scope that the install script can handle;
- **How to execute rustup-init? (e.g., on MS Windows)** Spawning a child process **(the GUI version!)** will bring many uncertainties to the install script, as its operation is independent of the terminal. Hence to cover these cases the install script can be much more cumbersome - an unworthy trade-off, IMO.

Jint-lzxy:
Thanks for the script! But the problem is we probably need to avoid installing multiple versions of Rust in parallel - so things become much more complex here, as Rust is distributed in multiple ways:
- **Self-hosted distribution** (some package managers e.g., [brew](https://formulae.brew.sh/formula/rust#default)): the package manager manages version control, `rustfmt` must be installed via `cargo`, but this has been deprecated; Besides, Rust is usually NOT installed on a standard path (aka `~/.cargo`).
- **The "legacy" official installer**: Similar to the official installer, except that `rustup` won't be installed (by default); `rustfmt` must be installed via `cargo`, or via `rustup` but `rustup` itself must be installed first.
- **The official installer, or the `rustup[-init]` package provided by many package mgrs**: The recommended approach.

So if we decide to automate this process, the following should be considered:
- **How to cleanup legacy installation(s)?** This seems to be far beyond the scope that the install script can handle;
- **How to execute rustup-init? (e.g., on MS Windows)** Spawning a child process **(the GUI version!)** will bring many uncertainties to the install script, as its operation is independent of the terminal. Hence to cover these cases the install script can be much more cumbersome - an unworthy trade-off, IMO.

CharlesChiuGit:
aha, I didn't thought about those details, indeed this is important.
But due to it's complexity, I think it's better not to include it in the installation script.
Just put it in prerequisite should be fine IMO.

Jint-lzxy:
IMO then we can just put a hint here, just like the `sqlite` one:
https://github.com/ayamir/nvimdots/blob/656e48a19732e06043e61ad54c28950980133313/scripts/install.ps1#L399-L402

CharlesChiuGit:
that would do too. upvote for this

Jint-lzxy:
Done :D https://github.com/ayamir/nvimdots/pull/852/commits/98e06787ba0c3b5506bfeb2653e973c13459f135
| Windows | \*nix |
|---------|-------|
| pwsh | zsh |

CharlesChiuGit (approved):
LGTM

Issue #866, opened 2023-07-15
Very slow scroll


fecet:
### Version confirmation

- [X] Confirm

### Following prerequisites

- [X] Confirm

### Neovim version

NVIM v0.9.1

### Operating system/version

Arch linux

### Terminal name/version

alacritty

### $TERM environment variable

xterm-256color

### Branch info

main (Default/Latest)

### Fetch Preferences

SSH (use_ssh = true)

### How to reproduce the issue

Open a relative large file (like 3000 loc) with `clangd` or `pyright`

### Expected behavior

Scroll is smooth as opened without plugin

### Additional information

![slow](https://github.com/ayamir/nvimdots/assets/41792945/e5b5f043-3056-481e-b656-ba0fe7b9b4dd)

 It appears to be due to an LSP  issue, as pylsp is significantly faster than pyright, but it may not be directly caused by pyright itself. This is because a similar phenomenon is observed on clangd as well. I suspect it is caused by a poorly supported plugin, but I haven't found the exact cause yet.

fecet:
https://github.com/HiPhish/rainbow-delimiters.nvim/issues/12

comment this give me huge speed up

fecet:
@HATTER-LONG do you still use this neovim config?

HATTER-LONG:
I am still using this configuration, but after updating the configuration from upstream, I always encounter strange  problems with some overly new plugins 😭

fecet:
I share the same feeling too. I hope there can be more testing when adding new plugins.

ayamir:
@HATTER-LONG Seems like you has solved it, can you open a pr for it?

HATTER-LONG:
ok, I will do this work later, I need to test how many error detections are suitable for most projects.

HATTER-LONG:
@fecet  
Oh, I just took a closer look at your question and the file you opened did not trigger `bigfile`, did it?
I just create PR only works in the case of large files 😓
Maybe you need try this `configs/editor/rainbow_delims.lua `：

```lua
	local function check_error_and_hgive_up()
		local parser = vim.treesitter.get_parser()
		local errors = 100
		parser:for_each_tree(function(tree)
			if tree:root():has_error() and errors >= 0 then
				errors = errors - 1
			else
				return nil
			end
		end)
		if errors < 0 then
			return nil
		end
		return require("rainbow-delimiters").strategy["global"]
	end

	vim.g.rainbow_delimiters = {
		strategy = {
[...]

PR #867, opened 2023-07-16
Fix the issue of slow file opening and operation due to too many tree sitter errors

HATTER-LONG:
Ref：https://github.com/HiPhish/rainbow-delimiters.nvim/issues/12

The repair method in the reference is to calculate the number of errors but I think it's better to use the same operation as the treesitter.

Please see if this is OK, at least It work well when I test it


HATTER-LONG:
https://github.com/ayamir/nvimdots/issues/866
maybe it cannot be repaired in this way

Issue #721, opened 2023-05-07
High CPU usage and exit issue of Jedi-language-server


Tgist1024:
### Version confirmation

- [X] Confirm

### Following prerequisites

- [X] Confirm

### Neovim version

NVIM v0.9.0 Build type: Release LuaJIT 2.1.0-beta3     system vimrc file: "$VIM/sysinit.vim"   fall-back for $VIM: "/usr/share/nvim"

### Operating system/version

Arch Linux 6.3.1-arch1-1

### Terminal name/version

kitty 0.28.1

### Branch info

main (Default/Latest)

### Fetch Preferences

SSH (use_ssh = true)

### Affected language servers

jedi-language-server

### How to reproduce the issue

1. Open a Python file or project.
2. Wait for lsp works.
3. Exit.
4. Repeat several times.
5. Open process panel tools like `top` and the problem shows.
![image](https://user-images.githubusercontent.com/101801875/236683767-be698526-7e18-47fb-902d-748c0ed1cf67.png)

### Actual behavior

The CPU usage of Jedi language server is extremely high, and it does not automatically exit when exiting neovim. If I repeatedly open and close Python files, multiple instances of Jedi language server will appear, causing significant CPU usage. As this issue affects user experience and computer performance, it is suggested to fix it.

### Support info

:LspInfo
```
 Language client log: /home/liu/.local/state/nvim/lsp.log
 Detected filetype:   python
 
 3 client(s) attached to this buffer: 
 
 Client: pyright (id: 1, bufnr: [1])
 	filetypes:       python
 	autostart:       true
 	root directory:  Running in single file mode.
 	cmd:             pyright-langserver --stdio
 
 Client: jedi_language_server (id: 2, bufnr: [1])
 	filetypes:       python
 	autostart:       true
 	root directory:  Running in single file mode.
 	cmd:             jedi-language-server
 
 Client: null-ls (id: 3, bufnr: [1])
 	filetypes:       python, cpp, c, css, scss, sh, markdown, javascriptreact, vue, typescript, javascript, typescriptreact, html, yaml, rust, vim, lua, luau
 	autostart:       false
 	root directory:  /home/liu/Documents/code/web/mysite
 	cmd:             <function>
 
 Configured servers list: jedi_language_server, rust_analyzer, lua_ls, dartls, clangd, html, bashls, pyright, jsonls
```

ayamir:
Does it have to do with the uninstalled `pyright`?

[...]

PR #868, opened 2023-07-16
perf!: consolidated updates to address performance issues

### LINKED ISSUE #866: Very slow scroll

Jint-lzxy:
This fixes #866 and closes #867.

cc @fecet

### DIFF
--- a/lua/modules/configs/editor/rainbow_delims.lua
+++ b/lua/modules/configs/editor/rainbow_delims.lua
@@ -1,7 +1,20 @@
 return function()
+	local function init_strategy()
+		local errors = 200
+		vim.treesitter.get_parser():for_each_tree(function(lt)
+			if lt:root():has_error() and errors >= 0 then
+				errors = errors - 1
+			end
+		end)
+		if errors < 0 then
+			return nil
+		end
+		return require("rainbow-delimiters").strategy["local"]
+	end
+
 	vim.g.rainbow_delimiters = {
 		strategy = {
-			[""] = require("rainbow-delimiters").strategy["local"],
+			[""] = init_strategy,
 		},
 		query = {
 			[""] = "rainbow-delimiters",
--- a/lua/modules/configs/editor/rainbow_delims.lua
+++ b/lua/modules/configs/editor/rainbow_delims.lua
@@ -1,20 +1,25 @@
 return function()
-	local function init_strategy()
-		local errors = 200
-		vim.treesitter.get_parser():for_each_tree(function(lt)
-			if lt:root():has_error() and errors >= 0 then
-				errors = errors - 1
+	local function init_strategy(check_lines)
+		return function()
+			local errors = 200
+			vim.treesitter.get_parser():for_each_tree(function(lt)
+				if lt:root():has_error() and errors >= 0 then
+					errors = errors - 1
+				end
+			end)
+			if errors < 0 then
+				return nil
 			end
-		end)
-		if errors < 0 then
-			return nil
+			return (check_lines and vim.fn.line("$") > 450) and require("rainbow-delimiters").strategy["global"]
+				or require("rainbow-delimiters").strategy["local"]
 		end
-		return require("rainbow-delimiters").strategy["local"]
 	end
 
 	vim.g.rainbow_delimiters = {
 		strategy = {
-			[""] = init_strategy,
+			[""] = init_strategy(false),
+			c = init_strategy(true),
+			cpp = init_strategy(true),
 		},
 		query = {
 			[""] = "rainbow-delimiters",
--- a/lua/modules/plugins/editor.lua
+++ b/lua/modules/plugins/editor.lua
@@ -36,11 +36,11 @@ editor["LunarVim/bigfile.nvim"] = {
 }
 editor["ojroques/nvim-bufdel"] = {
 	lazy = true,
-	event = "BufReadPost",
+	cmd = { "BufDel", "BufDelAll", "BufDelOthers" },
 }
 editor["rhysd/clever-f.vim"] = {
 	lazy = true,
-	event = { "BufReadPost", "BufAdd", "BufNewFile" },
+	event = { "CursorHold", "CursorHoldI" },
 	config = require("editor.cleverf"),
 }
 editor["numToStr/Comment.nvim"] = {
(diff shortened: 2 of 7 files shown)


ayamir:
Seems don't improve the startuptime too much for openning this file: https://github.com/STMicroelectronics/STM32CubeF4/blob/bb621289be5dd54f02e3198e1147b35aed4f7195/Drivers/CMSIS/Device/ST/STM32F4xx/Include/stm32f407xx.h
Before:
image
After:
image

fecet:
I also haven't observed any performance improvement, and I don't use scrollview and neoscroll. The performance issue should be unrelated to them.

Jint-lzxy:
> Sorry, I can't see the improvement of the scroll velocity when open [this file](https://github.com/STMicroelectronics/STM32CubeF4/blob/bb621289be5dd54f02e3198e1147b35aed4f7195/Drivers/CMSIS/Device/ST/STM32F4xx/Include/stm32f407xx.h).

Hmmm I somehow cannot reproduce this:
info

The current startup time on my side is close to the that of `nvim --clean` 🤔 

Perhaps explaining some background may be helpful? What this PR hopes to address is:
- **The startup time wrt opening files of normal sizes** - by deferring LSP setup to prevent them from interfering with startup of those UI components;
- **Open big files when `load_big_files_faster` is true** - Stop loading most plugins to simulate the behavior of native `nvim`.

If the startup time of any of the above situations is still slow, it may be because:
- (For 1) The server itself caused performance degradation, indirectly affecting UI rendering (similar to #721)
- (For 2) The parsers for `treesitter` are still being called... Hmm I'll look into this later. You can verify it this way:
```lua
:lua vim.treesitter.get_parser():parse()
```
This command should take a long time to complete.

ayamir:
Now it ok for me after `Lazy sync` 👍

ayamir (approved):

Jint-lzxy:
cc @fecet

Jint-lzxy:
I think it's good to go. Currently waiting for @fecet's response as he reported very limited performance improvements.

fecet:
Still don't find performance improvement unless set
https://github.com/ayamir/nvimdots/blob/1c02372f087c4b176564a83d8fc9eb7c375c5395/lua/modules/configs/editor/rainbow_delims.lua#L12
this to global.

lua/modules/plugins/editor.lua
```diff
@@ -36,11 +36,11 @@ editor["LunarVim/bigfile.nvim"] = {
 }
 editor["ojroques/nvim-bufdel"] = {
 	lazy = true,
-	event = "BufReadPost",
+	cmd = { "BufDel", "BufDelAll", "BufDelOthers" },
```
CharlesChiuGit, line 39: This needs to be loaded first before calling `BufDel` and the others.

lua/modules/configs/editor/rainbow_delims.lua
```diff
@@ -1,7 +1,20 @@
 return function()
+	local function init_strategy()
```
CharlesChiuGit, line 2: this function doesn't seems to work?

Jint-lzxy:
@fecet So u're using `treesitter` on large files? We don't make any gurantee regarding `treesitter`-related components under big files, as using `strategy["global"]` only reduces the number of times `LanguageTree:parse()` gets called, but each call could still leads to significant performance degradation. To verify this execute the following command urself:
```lua
:lua vim.treesitter.get_parser():parse()
```

This can only be resolved by updating the related parser(s), we cannot do anything. But yes we can use simpler processing methods on relatively larger files, a.k.a, `strategy["global"]`. Will push a fix for this.

lua/modules/plugins/editor.lua
Jint-lzxy: These commands are brought by `nvim-bufdel`, did I miss something?

Jint-lzxy: Try `:Lazy sync` first? There's no issue on my side.

Jint-lzxy:
> The file I tested is [`dwm.cpp` ](https://github.com/fecet/dwm/blob/master/dwm.cpp), it's not so large I guess? I don't really understand how ts works so cannot give further comment.

Thanks for the sample! Will do some testings there.

lua/modules/plugins/editor.lua
Jint-lzxy: Oh yes indeed, I found this corner case. Will revert this change.

Jint-lzxy: https://github.com/ayamir/nvimdots/pull/868/commits/d371fb4bafc10cf8477a0a551a1ebeffc3ca7b8a

Jint-lzxy:
> Still don't find performance improvement

I should have identified the problem. Now `rainbow-delimiters` will select an appropriate subtree when the cursor moves, but due to the issue with cpp parser, the generated AST becomes too tall, making one iteration too expensive.

CharlesChiuGit (approved):
LGTM

Jint-lzxy:
@fecet https://github.com/ayamir/nvimdots/pull/868/commits/ec24fef5485c1e0b19d463119dccf5b68ede15e4 should fix ur issue 👍

fecet:
OK, it's good for me now

lua/modules/plugins/editor.lua
CharlesChiuGit: @Jint-lzxy Sorry!!! 
I found it's actually a typo, `event = { "BufReadPost", "BufAdd", "BufNewFile"` totally works!

PR #871, opened 2023-07-17
workaroud for watchfile problem

fecet:
It turns out my issue https://github.com/ayamir/nvimdots/issues/866 is caused by two reasons. One is from rainbow and can be sloved by https://github.com/ayamir/nvimdots/pull/868, the other one is https://github.com/neovim/neovim/issues/23725 and have a workaroud https://github.com/neovim/neovim/issues/23725#issuecomment-1561364086

### DIFF
--- a/lua/core/init.lua
+++ b/lua/core/init.lua
@@ -151,6 +151,13 @@ local load_core = function()
 	neovide_config()
 	clipboard_config()
 	shell_config()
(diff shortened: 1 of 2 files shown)


Jint-lzxy (changes requested):
Please help fix the CI first.

Issue #886, opened 2023-07-22
NixOS support


misumisumi:
### Version confirmation

- [X] Confirm

### Following prerequisites

- [X] Confirm

### Neovim version

NVIM v0.9.1

### Operating system/version

NixOS 23.11.20230718.684c17c (Tapir)

### Terminal name/version

wezterm 20230712-072601-f4abf8fd

### $TERM environment variable

xterm-256color

### Branch info

main (Default/Latest)

### Fetch Preferences

SSH (use_ssh = true)

### How to reproduce the issue

If the python host prog is set on a system that does not support the Filesystem Hierarchy Standard (FHS), such as NixOS, dependencies related to python cannot be resolved.
This is because NixOS uses a python environment on NixOS that is isolated from the system python, and this environment is set via command line arguments, which are overridden by nvimdots.
A new configuration entry prevents nvimdots from overwriting it.

### Expected behavior

`:python3 import neovim`で正常にimportされる。

### Actual behavior

`:python3 import neovim`でエラーが発生する。

ayamir:
I have used NixOS before. In fact, There are tons of changes need to be done to make the configuration work for NixOS. For example, all of lsp servers installed by `Mason` won't work on NixOS because its special file structure.
This config is for the most GNU/Linux distribution which respects the LFS.
IMO, it's better to create a NixOS branch to manage stuffs related to NixOS if we really need to support it.

misumisumi:
You are right.
The binary installed with mason.nvim is basically unusable on NixOS, you have to use the language server installed with nix!
I once wrote a configuration for nvimdots for this project, but it requires significant changes, so I was thinking of doing my own support after separating the user settings as suggested in #885.
I can make a proposal for a NixOS-compatible configuration in the near future.

fecet:
I'm curious about nix, how can it work with conda?

misumisumi:
You can use conda on NixOS, but packages that use libraries that depend on other languages, such as Cython, have a high probability of not working properly.
This is because NixOS is not FHS compliant and cannot access libraries such as `/usr/lib/`.
You need to build an environment with dependencies resolved using `flake.nix` or `shell.nix`. (like Dockerfile).

misumisumi:
Also, if you are using the `home-manager` to manage neovim, the global python and the python used by neovim are managed separately.
The fold is the content of `nvim` in NixOS.
[...]

PR #887, opened 2023-07-22
fix(provider): leave provider specifications to user

misumisumi:
#886

### DIFF
--- a/lua/core/options.lua
+++ b/lua/core/options.lua
@@ -103,19 +103,22 @@ local function load_options()
 		wrapscan = true,
 		writebackup = false,
 	}
+
 	local function isempty(s)
 		return s == nil or s == ""
+	end
+	local function use_if_defined(val, fallback)
+		return val ~= nil and val or fallback
 	end
 
 	-- custom python provider
 	local conda_prefix = os.getenv("CONDA_PREFIX")
-    if require("core.settings").dont_set_python_host_prog then
-    elseif not isempty(conda_prefix) then
-		vim.g.python_host_prog = conda_prefix .. "/bin/python"
-		vim.g.python3_host_prog = conda_prefix .. "/bin/python"
+	if not isempty(conda_prefix) then
+		vim.g.python_host_prog = use_if_defined(vim.g.python_host_prog, conda_prefix .. "/bin/python")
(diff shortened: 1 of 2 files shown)


CharlesChiuGit:
the pr should done it's job, but i wonder if their's a better approach to it.

Jint-lzxy:
> but i wonder if their's a better approach to it

The workaround that crossed my mind is to allow users to define their custom Python provider(s) _(which may include command-line options)_. But I'm not sure if this is feasible cause I rarely use Python-related stuffs

misumisumi:
At least in NxOS, it is difficult to know the exact configuration of the latest neovim from outside of nix, since the product is managed as a hash.

fecet:
I use python everyday but I use conda to manage envs so just respect `which python` works fine for me.

Another approach is to respect the settings in `poetry` (or similar thing) like https://github.com/vsedov/nvim/blob/master/lua/modules/lsp/lsp/providers/python/pylance.lua

misumisumi:
The focus of the issue I was thinking about was about NixOS.
I did not consider other python environments such as poetry.
Therefore, I would like to raise the PullReqesut again, not on the issue of changing python_host_prog, but on the issue of NixOS support.
The title of the issue will be changed.

Jint-lzxy:
So to summarize, is it feasible to directly use the exec `python` provided by `vim.fn.exepath`?

misumisumi:
> So to summarize, is it feasible to directly use the exec python provided by vim.fn.exepath?

It can be used if it contains the necessary libraries for python such as global or `conda` or `venv`.
If not, NixOS requires `python(3)_host_prog` to be set correctly.
Because the `python` provided by `vim.fn.exepath` is user-wide python and does not contain the libraries neovim uses.
The environment containing the necessary libraries for neovim is provided by the command `nvim-python3`.

Jint-lzxy:
IMO in this case we can just leave it alone, cause if this option is `nil` Neovim would use the fallback (i.e., the one on `$PATH`) if it exists.

Jint-lzxy:
https://github.com/ayamir/nvimdots/pull/887/commits/fb17e95a323deba45850200d0f311f40d65cb19f

CharlesChiuGit (approved):
LGTM

ayamir (approved):

PR #892, opened 2023-07-23
feat(lualine): add icon for PyEnv

Jint-lzxy:
Chose the most expressive icon (IMO):

icon


fecet:
What's right with the original one? It's the same as https://github.com/romkatv/powerlevel10k

Jint-lzxy:
I just feel like using the same icon on both sides is a bit repetitive. Previously we used `nf-mdi-language_python_text` (`U+F820`), but it was deprecated with the update of material design icons.

ayamir (approved):

PR #898, opened 2023-07-27
perf(lspsaga): align options with official doc and update config

CharlesChiuGit:
I re-align the options with the order of https://dev.neovim.pro/lspsaga/ and update the config


Jint-lzxy (approved):
LGTM!

Issue #899, opened 2023-07-27
The mapping does not work as designed in debug mode


YuCao16:
### Version confirmation

- [x] Confirm

### Following prerequisites

- [X] Confirm

### Neovim version

NVIM v0.10.0-dev-618+g4fd852b8c

### Operating system/version

Pop!_OS 22.04 LTS x86_64

### Terminal name/version

kitty 0.28.1

### $TERM environment variable

xterm-kitty

### Branch info

main (Default/Latest)

### Fetch Preferences

SSH (use_ssh = true)

### How to reproduce the issue

1. Go to Debug mode
2. Press K in code buffer

### Expected behavior

K is bound to `lua require("dapui").eval()`

### Actual behavior

K is bound to `Lspsaga hover_doc'

### Additional information

In the most recent version, 'K' has been bound to lua require("dapui").eval(). However, in the main code window, this mapping has not been overwritten. There are two reasons for this:

1. Some LSP clients, such as copilot, are attached after entering debug mode.
2. LSP-related mappings are not overwritten. I need to specify the 'bufnr' for it to work.
 ```
	["n|K"] = map_cmd("<Cmd>lua require('dapui').eval()<CR>")
		:with_buffer(0)
		:with_noremap()
		:with_nowait()
		:with_desc("debug: Evaluate expression under cursor"),
```
I'm not sure if this is a real issue or a conflict between my own configuration and the current version. Could you please test it?

Jint-lzxy:
Cannot repro. We did test this before merging #877.

This is a global keymap, so it's impossible for a _specific_ buffer to have a different definition (unless overridden). And theoretically no LSP activity is allowed during this period. Can you provide a minimum repro so that we can look into this issue?

YuCao16:
> How about `K` in `visual mode`? does that works?

Yes, it works as expect in visual mode.

YuCao16:
> 

I'm not sure if this issue is caused by the problem of having multiple LSPs attached or something similar, since specifying the bufnr (:with_buffer(0)) seems to resolve it. In addition, the restoration method may not be comprehensive. The original K mapping is linked to those buffers with LSPs attached, but the restoration method does not specify this.

Jint-lzxy:
> can u confirm this part?

[...]

PR #903, opened 2023-07-27
fix(lua_ls): use `luarc.json` for server config

Jint-lzxy:
Ref: https://github.com/ayamir/nvimdots/issues/899#issuecomment-1653926046

### DIFF
--- a/.luarc.json
+++ b/.luarc.json
@@ -1,0 +1,22 @@
+{
+  "$schema": "https://raw.githubusercontent.com/sumneko/vscode-lua/master/setting/schema-zh-tw.json",
+  "diagnostics": {
+    "globals": ["vim"],
+    "disable": ["different-requires"]
+  },
+  "runtime": {
+    "version": "LuaJIT"
+  },
+  "workspace": {
+    "maxPreload": 100000,
+    "preloadFileSize": 10000,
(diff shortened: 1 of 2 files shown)


CharlesChiuGit:
Nice! Now only one `lua_ls` process is running!
圖片

Jint-lzxy:
Ah, I remember why I didn't port this upstream. The issue is in order to enable Neovim API completion, users now need to add the library themselves:
```lua
    "library": [
      "/path/to/share/nvim/runtime/lua",
      "/path/to/share/nvim/runtime/plugin"
    ]
```
See [lua-language-server/FAQ#why-are-there-two-workspacesprogress-bars](https://github.com/LuaLS/lua-language-server/wiki/FAQ#why-are-there-two-workspacesprogress-bars).

CharlesChiuGit:
hmmm, that's tricky one. Maybe we need to modify install scripts for this part?

Jint-lzxy:
> hmmm, that's tricky one. Maybe we need to modify install scripts for this part?

IMO We can just leave it as-is. Looks like the FAQ got updated recently - this is likely not a problem though, since `lua_ls` will actually index every file only once. It just creates an illusion that two workspaces have been setup.

CharlesChiuGit:
nice, that would save is some hustle!

Issue #185, opened 2022-08-13
[doc] Update wiki for catppuccin?


Jint-lzxy:
As mentioned in #163. I don't have write access to the wiki... so here's the issue instead 🥳 You can review these changes and pick the appropriate lines that could be added to the wiki.

~**NOTE: There's a new PR (#183) modifying some parts of the catppuccin configs, which means some links in the wiki need to be changed. There would be some changes when #183 is merged.**~ Now replaced.

## What's inside?
- Introduction
- How-to-use
  - Basic usage:
    - Customizing configs / integrations.
    - Enable the **auto-compile** feature _(default)_
  - Advanced usage:
    - Customizing palettes _(per flavour)_.
    - Customizing highlight.
    - Customizing **auto-compile** hook.

## Preview
<details>
    <summary>Raw markdown code <i>(add this to the wiki!)</i></summary>
<p>

`````````markdown
# Guide on how to use the catppuccin colorscheme
## What is Catppuccin? <sup><a href="https://www.reddit.com/r/neovim/comments/r9619t/comment/hna3hz8">[1]</a></sup>
> Catppuccin is a community-driven soothing pastel theme that aims to be the middle ground between low and high-contrast themes, providing a warm color palette with 26 eye-candy colors that are bright enough to be visible during the day, yet pale enough to be easy on your eyes throughout the night.

## Basic Usage
Modify [these lines](https://github.com/ayamir/nvimdots/blob/94f153471068c5145b2036be1f688c0f954cd973/lua/modules/ui/config.lua#L191-L267). _(Note: This link might be slightly different from `HEAD`, but it can be used as a reference.)_ See detailed explanation of each option below.

### General
These settings are unrelated to any group and are globally independent.
-   `vim.g.catppuccin_flavour`: _(Can be any one of: `latte`, `frappe`, `macchiato`, or `mocha`)_ this is **mandatory**. You **must** set this value in order to make catppuccin work correctly. Note that `latte` is a light colorscheme, and the rest are dark schemes; The `mocha` palette is the only one that has been modified to make catppuccin look like the v0.1 one. Check out [this PR](https://github.com/ayamir/nvimdots/pull/163) for details.
-   `transparent_background`: _(Boolean)_ if true, disables setting the background color.
-   `term_colors`: _(Boolean)_ if true, sets terminal colors (a.k.a., `g:terminal_color_0`).

### Dim inactive
This setting manages the ability to dim **inactive** splits/windows/buffers.
[...]

PR #906, opened 2023-07-28
NixOS support

misumisumi:
- [x] `TSInstall all` installs and builds all `treesitter`
- [x] Binaries installed with `mason.nvim` work
  - [x] checked `lua-language-server`, `pylsp`, `marksman`.
  - [x] ~~Need to check for all binary packages? (maybe checking their dependencies instead, see below)~~
- [x] Can build from sources installed with `mason.nvim`
  - [x] Some packages have no upstream problem or identification of the cause. [detail](https://github.com/ayamir/nvimdots/issues/886#issuecomment-1650866181)
- [x] wiki needs NixOS support description


nixos/neovim/default.nix
```diff
@@ -0,0 +1,277 @@
...
+          withRuby = true;
+
+          extraPackages = with pkgs;
+            [
+              # Dependent packages used by default plugins
+              doq
+              neovim-remote
+              ripgrep
+              sqlite
+              xclip
```
vollowx, line 225: Not everyone are using desktop under X11, maybe it should be 2 options `withWaylandClipboard (boolean)` and `withX11Clipboard (boolean)` or `withClipboard (string: 'wayland'|'x11')`.
Besides, most of the desktop users have already installed the related clipboard tools to their desktop as NeoVim is not the only program which need its support, so maybe it's also not necessary to install them at here, inside a NeoVim configuration.

Jint-lzxy:
>  Need to check for all binary packages?

IMO it's unnecessary to check all those packages, it takes too much time as well as ur disk space. After https://github.com/ayamir/nvimdots/issues/886#issuecomment-1649299481, I installed all those packages on an external disk and dumped their dependencies using `otool`. I just did some `sed`-, `sort`- and `uniq`ing, and these are the dependencies required by existing Mason packages (excluding dependencies managed by those "install mediums"):
```console
libamd     <==> https://github.com/zship/libamd
libc       <==> https://www.gnu.org/software/libc/
  [OR] libc++ <==> https://libcxx.llvm.org/
libevent   <==> https://libevent.org/
libiconv   <==> https://www.gnu.org/software/libiconv/
libmsgpack <==> http://msgpack.org/
libz       <==> https://www.zlib.net/
libzstd    <==> https://github.com/facebook/zstd
```
FWIW this is actually similar to the compiling Treesitter's parsers, as long as all those dependencies are satisfied (and the dynamic linker is properly initialized), these executables should run with ease :D

nixos/neovim/default.nix
misumisumi: Thank you for your review.
It is certainly so. It contained settings tailored to my environment.
As you can see from [doc]() there are many options, so I won't support them here.
Note about the clipboard should be documented in the wiki.

misumisumi:
> IMO it's unnecessary to check all those packages, it takes too much time as well as ur disk space.
I understand remove from task

At least the LSP and Debugger, which are installed by default, have been built and verified.
We've made it possible to optionally include dependencies for the behavior of other packages, so you should be able to investigate and respond to them yourself.
It is enough to describe the method of investigation on the wiki.

Jint-lzxy:
> We've made it possible to optionally include dependencies for the behavior of other packages, so you should be able to investigate and respond to them yourself.
> It is enough to describe the method of investigation on the wiki.

I think the same applies to "Some packages have no upstream problem or identification of the cause" too. If one really needs these individual builds, he can further tweak the base config to make that work (and usually they have such environment setup so fixing this won't be that difficult).

misumisumi:
I understand Mark as completed.
Where should I submit a PR for the wiki description?
I'll post the base here if only admins can edit it.

ayamir:
> I'll post the base here if only admins can edit it.

You can edit it too.

Jint-lzxy:
> Where should I submit a PR for the wiki description?
> I'll post the base here if only admins can edit it.

I remember last year we did tweak those wiki pages so that everyone can edit them. But if ur patch contains a lot of additions, it would be better to open an issue where we could do some proofreading before adding this (like #185).

misumisumi:
It's built as a `home-manager` module, so it should work.
I couldn't try it because I didn't create a config just for `home-manager`.
If there is a problem, an issue will appear, so I will deal with it at that time.
Remove it from my task.

ayamir:
OK, but I don't have `NixOS` env, can't review if it works.

CharlesChiuGit:
me too

vollowx:
I probably can review as I use NixOS, I will go home from vacation the day after tomorrow, then I can use my computer.

aarnphm:
I'm currently using nix-darwin for my setup, but I can take a look since this is mostly flake-based @misumisumi ?

nixos/neovim/default.nix
```diff
@@ -0,0 +1,275 @@
...
+        '';
+        setBuildEnv = mkEnableOption ''
+          Sets environment variables that resolve build dependencies needed by `mason.nvim` and `nvim-treesitter`
+          Environment variables are only visible to `nvim` and have no effect on the session.
+          Required for NixOS.
+        '';
+        withBuildTools = mkEnableOption ''
+          Include basic build tools like `gcc` and `pkg-config`.
+          Required for NixOS.
+        '';
```
aarnphm, line 26: Hmm I'm thinking to have each `with` option is a bit overkill. Especially with most of these providers, we can introduce them gradually. What do you think?

misumisumi:
yes.
Since this was created as a module for `home-manager`, it can also be used in `nix-darwin` as long as general user settings are configured with `home-manager`.

nixos/default.nix
```diff
@@ -0,0 +1,6 @@
+{
```
aarnphm, line 1:
```suggestion
# NOTE: to add more language support, make a directory under `nixos`, followed by the language name and `default.nix`. See `dotnet/default.nix` for example.
{
```

nit: add small suggestion for new ppl who might want to contribute to nix.

aarnphm (commented):
would be cool to also make this flake.nix compatible on darwin as well.

nixos/neovim/default.nix
misumisumi: Other than go, haskell, R, and vala, it's just installation of runtime and package manager, so I think it's okay if you don't have it.
I think go and vala should be left because the `masn.nvim` package requires non-runtime dependencies.
Haskell and R are difficult for beginners to include dependencies, so I think it's better to leave it as it is.

aarnphm:
> yes. Since this was created as a module for `home-manager`, it can also be used in `nix-darwin` as long as general user settings are configured with `home-manager`.

need to migrate my setup to `home-manager` 😄

nixos/neovim/default.nix
```diff
@@ -0,0 +1,247 @@
...
+        withDotNET = mkEnableOption ''
+          Enable dotnet provider. Set to `true` to
+          use DotNET plugins.
+        '';
+        withGo = mkEnableOption ''
+          Enable Go provider. Set to `true` to
+          use Go plugins.
+        '';
```
aarnphm, line 34: I'm still a bit iffy about this. How about just having an option that allows users to just add additional dependencies from nix for the provider they want? My feeling is that installing default packages might not fit everyone.

aarnphm: Additionally, this will become a lot harder to maintain once we add more and more language

misumisumi: Adding the necessary dependencies for package building in `mason.nvim` is already implemented with `extraDependentPackages`. (This resolves `lib`, `include` and `pkg-configs`.)
The remaining languages are the ones that needed dependencies when building and installing all the packages that can be installed with `mason.nvim` when this was created.
However, considering the cost of tracking packages added to `mason.nvim` in the future, I think it's fine to leave only `extraDependentPackages` as you said.
IMO, leave Haskell and PHP and R alone where dependency resolution is complicated for nix newbies.
Currently, the dependencies that I know are described in the wiki.
Or write an example on the wiki how to resolve dependencies for PHP, R as well.
For Haskell, `stack` must be included in the form of a shell script with options, so it must be left.

aarnphm: yep stack should be set for haskell, but I think for all of the other dependencies it is ok to have an example on other language.

ayamir:
Can it be merged?

misumisumi:
Yeah.
After checking at hand, you can merge to `main` without conflict.

nixos/neovim/default.nix
```diff
@@ -0,0 +1,182 @@
...
+        };
+        home.packages = with pkgs; [
+          ripgrep
+        ] ++ optionals cfg.setBuildEnv [ patchelf nvim-depends-library nvim-depends-include nvim-depends-pkgconfig ];
+        home.extraOutputsToInstall = optional cfg.setBuildEnv "nvim-depends";
+        home.shellAliases.nvim = optionalString cfg.setBuildEnv (concatStringsSep " " buildEnv) + " SQLITE_CLIB_PATH=${pkgs.sqlite.out}/lib/libsqlite3.so " + "nvim";
+
+        programs.neovim = {
+          enable = true;
+          viAlias = true; # Replace from vi&vim to neovim
```
aarnphm, line 136: This is just a preference, but I still prefer vi -> vim, and vim -> neovim
```suggestion
          viAlias = true; # Replace from vi&vim to neovim
```

aarnphm (approved):
Oh I thought I submitted my review 😄 LGTM

nixos/neovim/default.nix
misumisumi: ah. It was my preferred setting.
As for Alias, I think it would be better if the user could select it, so I will remove it.

Jint-lzxy:
I'll merge this after updating the docs :-) Thanks again for all the work here!

nixos/neovim/default.nix
aarnphm: Ah probably we can comment this out and leave it in such that ppl now that these exist.

Sometimes nix documentation can be hard to navigate 😢

Jint-lzxy (approved):
LGTM! I will merge this if u think the [Wiki Page](https://github.com/ayamir/nvimdots/wiki/NixOS-Support) is good (I made some minor modifications) 😄

cc @misumisumi @vollowx

vollowx:
The link [here](https://github.com/ayamir/nvimdots/wiki/NixOS-Support#customize-your-experience-available-options) should be updated as it may change in the future. Others are enough in my use case. :+1:

Jint-lzxy:
> The link [here](https://github.com/ayamir/nvimdots/wiki/NixOS-Support#customize-your-experience-available-options) should be updated as it may change in the future.

I will update this (using permalinks) after merging this PR, cause we currently don't have the stub for this file in the source tree :D

misumisumi:
LGTM😄

PR #913, opened 2023-07-30
fix toggleterm & add config for diffview.nvim

YuCao16:
1. `toggleterm.nvim`: fix toggleterm `start_in_insert` not working.
2. `diffview.nvim`: disable lsp diagnostics in Diffview mode, otherwise, lsp goes crazy due to the conflict markers (<<<) and then crashes.

### DIFF
--- a/lua/modules/configs/editor/diffview.lua
+++ b/lua/modules/configs/editor/diffview.lua
@@ -1,0 +1,9 @@
+return function()
+	require("diffview").setup({
+		view = {
+			merge_tool = {
+				disable_diagnostics = true,
+			},
+		},
+	})
+end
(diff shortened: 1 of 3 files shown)


ayamir:
1. Can't repro on my wsl.
2. Approve.

YuCao16:
> 1. Can't repro on my wsl.
> 2. What you introduced is set in the default setup function as `diffview.nvim`'s readme said.

1. Are you using 0.10, because I'm gussing it's a nightly issue for autocmd. 

2. I ran into this problem yesterday and I thought it wasn't turned on by default, sorry about this. I will try to create a repro code and raise issue in diffview.

ayamir:
No, I tested it for 0.9. You should open a pr for 0.10 branch instead of main.

PR #916, opened 2023-07-31
fix: only disable autoclose for specific filetypes

ayamir:
1. disable autoclose `''` for **rust** because `'` is the symbol for lifetimes.
2. disable autoclose `<>` for **cpp only** because `<<` and `>>` are operators in cpp but `<>` is often used as generic type.

### DIFF
--- a/lua/modules/configs/editor/autoclose.lua
+++ b/lua/modules/configs/editor/autoclose.lua
@@ -5,15 +5,15 @@ return function()
 			["["] = { escape = false, close = true, pair = "[]" },
 			["{"] = { escape = false, close = true, pair = "{}" },
 
-			["<"] = { escape = true, close = true, pair = "<>", disabled_filetypes = { "cpp" } },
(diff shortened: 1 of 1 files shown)


Jint-lzxy (approved):
LGTM

fecet:
Dosn't most languages use `<` and `>` for comparasion? I guess enable it only for specific filetypes would be better

ayamir:
> Dosn't most languages use `<` and `>` for comparasion? I guess enable it only for specific filetypes would be better

You're right. But `autoclose.nvim` doesn't have the option to **enable** it for the specific file type. I have opened a issue for it [here](https://github.com/m4xshen/autoclose.nvim/issues/34).

Jint-lzxy:
> You're right. But `autoclose.nvim` doesn't have the option to **enable** it for the specific file type. I have opened an issue for it [here](https://github.com/m4xshen/autoclose.nvim/issues/34).

@ayamir IMO we can revert this change atm to avoid regression & repeated issues.

PR #919, opened 2023-08-02
fix(toggleterm): `start_in_insert` not working

Jint-lzxy:
IMO this is a good patch (taken from #913) and should be merged.


YuCao16:
I initially assumed this to be an issue with the nightly version, but that's not the case. The start_in_insert doesn't work because persist_mode is set to true. persist_mode retains the mode from when you last exited toggleterm.nvim. If you exit the terminal using wincmd, then you would reenter in insert mode. In my case, I exit the terminal by pressing q after double-tapping <ESC>. Consequently, toggleterm remembers that my previous mode was normal, and it sets the mode back to normal when I reenter toggleterm.

`toggleterm.nvim` Notice: "If using persist_mode terminal mappings should be changed to use wincmd instead otherwise persist mode will not work correctly."

PR #921, opened 2023-08-02
feat(plugins): add `symbols-outline.nvim`

Jint-lzxy:
https://github.com/ayamir/nvimdots/discussions/494#discussioncomment-6585631

PR #920, opened 2023-08-02
feat(plugins): add `Glance.nvim`

Jint-lzxy:
https://github.com/ayamir/nvimdots/discussions/494#discussioncomment-6585631

PR #926, opened 2023-08-02
feat(plugins): add `flash.nvim` to replace `clever-f.vim`

YuCao16:
https://github.com/ayamir/nvimdots/discussions/494#discussioncomment-6616409
1. In `core/mapping.lua`, I wrote a function for users who don't have `flash.nvim` installed.
2. I only bind partial keymap available in `flash.nvim` in `keymap/editor.lua`, because I use the default `s` quite often.


YuCao16:
I'm not sure If I should remove `cleverf`, because it involves delete&replace it's keymap.

YuCao16:
https://github.com/ayamir/nvimdots/pull/926/commits/21748b22826a0d381ad0ea8749eb13b507d5aea3: remove `clever-f` should work, it's keymap seems equal to `flash.nvim's` default. Do we need to define `,` and `;` explicitly?

fecet:
What's the reason to add this, it doesn't do more than hop and treehopper

YuCao16:
> What's the reason to add this, it doesn't do more than hop and treehopper

You can check out it's [README.md](https://github.com/folke/flash.nvim). In short, it combines all three plugins (include `clever-f`) and enhances `/`. Isn't it worth it? :D

YuCao16:
> 
I agree, but I'm only using some of the basic features of comments.nvim. Can you explain how it conflicts with `flash.nvim`? Because it seems to me that I only defined `S`, and `clever-f` also defines `f` and `F`.

YuCao16:
> I support to use flash to replace clever-f and then users can freely choose whether to use it as a replacement for hop.

They had the advantage over each other and I didn't remove `hop.nvim`, even though I use `flash.nvim` more often.

Jint-lzxy:
> I support to use flash to replace clever-f and then users can freely choose whether to use it as a replacement for hop.

Hmm FWIW this way the drawbacks of this plugin outweigh the potential benefits it may bring. I mean, why do we need to replace a single working plugin with a highly-integrated one? We can add [wildfire.vim](https://github.com/gcmt/wildfire.vim) for custom textobjects and have each plugin do their respective functionality, which will also reduce maintenance pressure as no plugin will experience way faster iterations.

> And https://github.com/ayamir/nvimdots/discussions/494#discussioncomment-6260886 cmp also enhance /

Ah thanks for providing a working example! I'm working on this, trying to cover some cornercases 😄

YuCao16:
> I support to use flash to replace clever-f and then users can freely choose whether to use it as a replacement for hop.
> 
> IMO hop, tree-hopper and incremental-selection covers the functionalities of flash very well and they are already in this repo. I'm also implementing [folke/flash.nvim#52](https://github.com/folke/flash.nvim/issues/52) on the top of TS.
> 
> And [#494 (comment)](https://github.com/ayamir/nvimdots/discussions/494#discussioncomment-6260886) cmp also enhance /

This is a little bit different IMO, `flash.nvim` also support fast jump in search mode, however, cmp and other plugins don't.

YuCao16:
> > I support to use flash to replace clever-f and then users can freely choose whether to use it as a replacement for hop.
> 
> Hmm FWIW this way the drawbacks of this plugin outweigh the potential benefits it may bring. I mean, why do we need to replace a single working plugin with a highly-integrated one? We can add [wildfire.vim](https://github.com/gcmt/wildfire.vim) for custom textobjects and have each plugin do their respective functionality, which will also reduce maintenance pressure as no plugin will experience way faster iterations.
> 
> > And [#494 (comment)](https://github.com/ayamir/nvimdots/discussions/494#discussioncomment-6260886) cmp also enhance /
> 
> Ah thanks for providing a working example! I'm working on this, trying to cover some cornercases smile

I've been using `cmp-cmdline`, because sometimes lsp integrate is needed in cmdline. but `wilder.nvim's` statusline is quite useful ...

Jint-lzxy:
> This is a little bit different IMO, `flash.nvim` also support fast jump in search mode, however, cmp and other plugins don't.

Maybe we can only enable this specific feature? b/c we'll probably load `flash.nvim` at `{ "CursorHold", "CursorHoldI" }`, this way, its setup will unlikely to result in performance overhead.

Jint-lzxy:
> but `wilder.nvim`'s statusline is quite useful ...

Yeah I'm trying to address this issue (if we can collect and import these missing items using external sources)

YuCao16:
> Maybe we can only enable this specific feature?

You mean `f`, `F` and `/` or only `f` and `F`?

YuCao16:
> > You mean `f`, `F` and `/` or only `f` and `F`?
> 
> I mean we can only enable "fast jump in search mode". This way we can avoid introduing breaking changes - if users do want to activate some of `flash`'s features, they can enable it themselves. To me, I tend to distribute functionalities to as many plugins as possible, as this reduces maintenance pressure by eliminating the need to frequently update a specific config (file). But this is just a matter of taste :) If a plugin can bring about better experience then, why not?

I agree, as if we should replace `hop.nvim` by `flash.nvim` is not my concern, everyone has their own preferences. For introducing this plugin:
1. `clever-f` has some bug, sometimes highlight won't clear.
2. `flash.nvim` offers fast jump in search mode.

aarnphm:
I tend to agree with @Jint-lzxy, maybe we can introduce flash.nvim gradually, first for just `f`, and `F`. For search mode I think telescope should be more than enough

YuCao16:
I feel like I'm the author of 'flash.nvim' :D, but seems telescope can't do this.

Jint-lzxy:
> `clever-f` has some bug, sometimes highlight won't clear.

Hmm I haven't encountered this issue before, but `clever-f` [does clear its highlight after some time](https://github.com/ayamir/nvimdots/blob/eb588cd149af05bd91fcd8ead8de39fa7cfb70ed/lua/modules/configs/editor/cleverf.lua#L10). Is this related to ur issue?

> I feel like I'm the author of 'flash.nvim' :D, but seems telescope can't do this.

Yeah that's why I said "we can only enable 'fast jump in search mode'"

YuCao16:
> Hmm I haven't encountered this issue before, but clever-f [does clear its highlight after some time](https://github.com/ayamir/nvimdots/blob/eb588cd149af05bd91fcd8ead8de39fa7cfb70ed/lua/modules/configs/editor/cleverf.lua#L10). Is this related to ur issue?

I kind of forgot how to reproduce that bug, if I finish the JUMP then there is no highlight at that point. But when I press `;` again (without pressing `f`), the highlight doesn't get cleared.

vollowx:
Another thing, as flash takes place of statusline, and it's prompt is configurable, we can make it look like lualine section a, maybe better.

YuCao16:
> Another thing, as flash takes place of statusline, and it's prompt is configurable, we can make it look like lualine section a, maybe better.

Good point, I will take a look in the morning.

CharlesChiuGit:
Before I read the readme of `flash.nvim`, I was leaning toward Jint's side, which is distribute functionalities to each plugins, since we already had the experience of `lspsaga`.

But after reading the readme, I found it was THE motion plugin I tried to find an year ago! If `flash.nvim` was created at that time, I think I will go all in on this.

Considering there might be lots of ppl using the repo, I think it would be better to replace one functionality/plugin at one release.

Not sure about the status line thingy tho, I didn't look super deep into `flash.nvim`.

Jint-lzxy:
> Considering there might be lots of ppl using the repo, I think it would be better to replace one functionality/plugin at one release.

Also quite in favor of this (as per https://github.com/ayamir/nvimdots/pull/926#issuecomment-1662666094) - we can gradually complete such transition like what we did in #920 and #921. Cause from the user's perspective, those individual PRs only _adds_ a few files, along with several trivial (and "conflict-friendly") changes that are easy to adopt. This can also effectively avoid breaking changes (which can be quite annoying in some cases).

ayamir:
Tried it but seems it has a certain learning curve compared to `hop.nvim`.

YuCao16:
> Tried it but seems it has a certain learning curve compared to `hop.nvim`.

True, there is no learning curve for `hop.nvim`. In addition, I think `flash.nvim` has some behaviors are a bit weird, like https://github.com/folke/flash.nvim/issues/197#issue-1835744177.

fecet:
For anyone who might be interested, I've created a [plugin](https://github.com/SUSTech-data/wildfire.nvim) to replace wildfire.vim.

CharlesChiuGit:
@fecet wrong link direct to wildfire?

fecet:
It's wildfire.[n]vim so that's correcct :laughing:

ayamir:
Maybe it's time to promote this PR @YuCao16 ?

YuCao16:
> Maybe it's time to promote this PR @YuCao16 ?

Yes, give me sometime.

YuCao16:
Sorry for the long delay, what other changes do I need to make before merge?

ayamir:
IMO the `jump` function is enough to replace `HopWord`. But other functions like search and jump can't replace `wilder.nvim` because it doesn't support file level. Its `remote` function is wired for me and I can't understand its value.

ayamir:
IMO it's better to leave the decision to users. It looks good to me now.

ayamir (approved):

lua/core/mapping.lua
```diff
@@ -2,6 +2,18 @@ local bind = require("keymap.bind")
...
+local flash_esc = function()
+	local status_ok, state = pcall(function()
+		return require("flash.plugins.char").state
+	end)
+	if status_ok and state then
+		state:hide()
+	else
+		vim.cmd("noh")
+	end
+end
```
Jint-lzxy, line 16: This can be moved to `lua/keymap/helpers.lua` as a global helper function.

Jint-lzxy (changes requested):

lua/core/mapping.lua
YuCao16: @Jint-lzxy Done.

Jint-lzxy (approved):
Sorry for the delay. This PR LGTM now 😸 I explicitly disabled `flash.nvim`'s search mode integration (but kept the related keymap to facilitate user customization), b/c IMHO this may not be the point that some users expect to change, and there're compatibility issues with `wilder.nvim` (correct me if I'm wrong :D)

YuCao16:
> Sorry for the delay. This PR LGTM now 😸 I explicitly disabled `flash.nvim`'s search mode integration (but kept the related keymap to facilitate user customization), b/c IMHO this may not be the point that some users expect to change, and there're compatibility issues with `wilder.nvim` (correct me if I'm wrong :D)

I agree, LGTM, thanks for your help!

Issue #923, opened 2023-08-02
Highlighted word is the previous candidate word when using fuzzy search


lzyuz:
### Version confirmation

- [X] Confirm

### Following prerequisites

- [X] Confirm

### Neovim version

NVIM v0.9.1

### Operating system/version

ubuntu22.04

### Terminal name/version

kitty0.29

### Branch info

main (Default/Latest)

### Fetch Preferences

SSH (use_ssh = true)

### How to reproduce the issue

no

### Expected behavior

Highlighted word should be candidate word

### Actual behavior

![image](https://github.com/ayamir/nvimdots/assets/127972121/eb29ba3c-d097-4081-943f-864948e0bc5c)
![image](https://github.com/ayamir/nvimdots/assets/127972121/a9135765-2c67-4bd9-a45c-e60cce472493)

fecet:
try 
![image](https://github.com/ayamir/nvimdots/assets/41792945/41ccf603-549e-4151-a9ff-24ce8cef416b)

I fixed this long time ago and don't use wilder now

CharlesChiuGit:
Can reproduce.

lzyuz:
> try ![image](https://user-images.githubusercontent.com/41792945/258022494-41ccf603-549e-4151-a9ff-24ce8cef416b.png)
> 
> I fixed this long time ago (forget how) and don't use wilder now so not sure if it works

sovled, many thanks

PR #929, opened 2023-08-03
feat: replace wilder by cmp.cmdline

fecet:
https://github.com/ayamir/nvimdots/discussions/494#discussioncomment-6618808

### DIFF
--- a/lua/modules/configs/completion/cmp.lua
+++ b/lua/modules/configs/completion/cmp.lua
@@ -162,4 +162,45 @@ return function()
 			},
 		},
 	})
+	local cmdline_maps = vim.tbl_deep_extend("force", cmp.mapping.preset.cmdline(), {
+		["<Down>"] = {
+			c = function()
+				if cmp.visible() then
+					cmp.select_next_item()
+				else
+					cmp.complete()
+				end
+			end,
+		},
+		["<Up>"] = {
+			c = function()
+				if cmp.visible() then
+					cmp.select_prev_item()
+				else
+					cmp.complete()
+				end
+			end,
+		},
+	})
+	-- `/` cmdline setup.
+	cmp.setup.cmdline("/", {
+		mapping = cmdline_maps,
+		sources = {
+			{ name = "buffer" },
+		},
+	})
+	-- `:` cmdline setup.
+	cmp.setup.cmdline(":", {
+		mapping = cmdline_maps,
+		sources = cmp.config.sources({
+			{ name = "path" },
+		}, {
+			{
+				name = "cmdline",
+				option = {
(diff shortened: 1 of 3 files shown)


Jint-lzxy:
> I haven't used wilder for a while, what makes statusline of wilder special

Wilder integrates better with "stuffs outside Neovim" - e.g., `cmp.cmdline` doesn't provide completions for files under cwd, `:Man` pages, shellcmds, etc. But what makes me question the potential benefits this PR may bring is closely related to how nvim-cmp sorts its entries:
https://github.com/ayamir/nvimdots/blob/8325140dead9f4eb7d33960ca25ae291e3b9142a/lua/modules/configs/completion/cmp.lua#L49-L68

\- We didn't take `cmp-cmdline` into consideration while adjusting `sorting.comparators` (which will also be used to filter/sort those cmdline completions) - IMO calling those whole bunch of comparators will be a complete overkill; This implies to support `cmp.cmdline` without causing performance overhead, we must change those sorting criteria exclusively for `cmp-cmdline`; the formatting of these entries also needs to be changed, as meta information about those slots no longer becomes relevant and could be misleading. FWIW it would be better to have a plugin focus on "what it's mainly designed for", what's the motivation behind this change?

fecet:
I am not quite sure how the performance of cmp is affected, and I have gotten used to its slowness (due to the inherent issues of Python LSP and Copilot, it's unlikely to be fast). Are you referring to the completion of :Man and :! in command mode? For file paths, it seems to work.

I just cherrypick something in my config in case someone else may found useful  :)

Jint-lzxy:
> Are you referring to the completion of :Man and :! in command mode? For file paths, it seems to work.

Ys. But there're several corner cases for file paths, e.g., too many nested dir changes like `../../.././../`. This is b/c `cmp.cmdline` itself does not perform any sort of analysis, it's just a wrapper around Vim's default completion engine (which means it "inherits" all the problems of omnicomp). 

> I am not quite sure how the performance of cmp is affected, and I have gotten used to its slowness

I have been experiencing input lags while trying to specialize configs for `cmp.cmdline` these days (which was not the case before). After I removed `cmp.cmdline`, everything went fine - so I doubt something must be wrong there.

> I just cherrypick something in my config in case someone else may found useful :)

Yeah it's just a matter of taste & perference, and Neovim is (as always) supposed to be a PDE, not an IDE 😄 Just curious, is there any issue regaring `wilder.nvim` that motivated u to make this change?

fecet:
can't really remember maybe just cause it's dead/written by vim/have to fixed stuffs like https://github.com/ayamir/nvimdots/issues/923

Issue #888, opened 2023-07-22
 About improvement of user customization experience.


misumisumi:
### Feature description

This is a suggestion regarding #885.
It contains major changes across the board.
Other suggestions welcome..

I'm looking for a way that doesn't change the original configuration too much.
Work is done at [misumisumi/nvimdots/add-user-custom-function](https://github.com/misumisumi/nvimdots/tree/add-user-custom-function).

- [x] Implementing the `load_plugin` function
  - [x] Extend default value when given value is a table
  - [x] Replace completely when the given value is a function
- [x] Support override for `event.lua` 
- [x] Support override for `options.lua` 
- [x] Support override for `settings.lua` 
- [x] Support override for `keymap.lua`

misumisumi:
https://github.com/ayamir/nvimdots/pull/885#pullrequestreview-1542119001 ccording to A, an override that does not rely on LazyNvim is likely necessary.
-[] Consider using the merge function or better

fecet:
Maybe we can take a look at https://github.com/LazyVim/LazyVim/tree/main, imitate its behavior or develop based on it.

Jint-lzxy:
> Maybe we can take a look at https://github.com/LazyVim/LazyVim/tree/main, imitate its behavior or develop based on it.

Hmm I don't feel like things are gonna work this way. Cause compared to `LazyVim`, we used a completely different loading strategy & directory hierarchy (implemented in #458, that's why we [added a custom loader](https://github.com/ayamir/nvimdots/blob/main/lua/core/pack.lua) instead of using `lazy.nvim`'s). Following `LazyVim` will definitely require a complete rewrite - but (at least in the current state) we don't want this config to become some sort of "Neovim distribution", where the base config and user config are completely separated (or are connected via settings entries) - which means incremental updates to plugins' specs are almost impossible (well, this is the case if the user did follow the instructions to customize their experience). See https://github.com/ayamir/nvimdots/pull/885#pullrequestreview-1542119001 for some thoughts on this.

IMO this is not _that_ viable regarding our current situation, but we can further discuss its feasibility as well 😄

[Quoting](https://github.com/ayamir/nvimdots/issues/306#issuecomment-1293552965) @CharlesChiuGit:

> I once tried Lunarvim, but it also has it's own issue. One day u will still need to know how to fix bugs like this.
The donw side of a highly pre-config nvimdot is that u can't be sure about wether the bug is from lunarvim, plugins or nvim itself.

cc @fecet

[...]

PR #930, opened 2023-08-06
feat: add setting for copilot enable or not.

ayamir:
Not all users use copilot and not all situations we need it. So I add a switch to it. (Sorry for the wrong branch name)

### DIFF
--- a/lua/core/settings.lua
+++ b/lua/core/settings.lua
@@ -12,6 +12,10 @@ settings["format_on_save"] = true
 ---@type boolean
 settings["format_notify"] = true
 
+-- Set it to false if you don't use copilot
+---@type boolean
+settings["use_copilot"] = false
+
 -- Set it to false if diagnostics virtual text is annoying.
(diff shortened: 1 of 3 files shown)


CharlesChiuGit:
hmmm, it would make a mess if we include `tabline` and `codeium`. maybe just leave it along.

fecet:
maybe also good to add some other AI stuff like chatgpt.nvim

Jint-lzxy (approved):
IMO we can leave this PR unchanged cause copilot has a large user base, thus shipping copilot with our base config makes sense to me. For those other plugins (possible candidates for extras), @fecet proposed in https://github.com/ayamir/nvimdots/issues/888#issuecomment-1649739665 that we can "use our custom loader (to) import extras like LazyVim". FWIW this is just trivial amount of work and can be implemented together w/ #931 😄

ayamir:
Making it `true` by default won't influence the current user experience.

ayamir:
> maybe also good to add some other AI stuff like chatgpt.nvim

Haven't tried it because my free api key expired before lol

PR #931, opened 2023-08-06
refactor (user custom): Proposal for enhancing user customization experience

misumisumi:
I proposed user custom isolation in #885, but changed many parts because it was an override using `opts` in `lazy.nvim`.
In #888, it was pointed out that the default definition might be read after the user definition if the `lazy.nvim` method is passed to `opts`.
In addition, it was pointed out that it would be possible to consider a configuration that can be overridden without making major changes to the current configuration, so we considered it again and implemented it.
At that time, I created a new branch, so I will throw a PR again.
Closing the previous PR.


lua/core/event.lua
```diff
@@ -151,7 +151,10 @@ function autocmd.load_autocmds()
 			},
 		},
 	}
-
+	local ok, user_definitions = pcall(require, "user.event")
```
Jint-lzxy, line 154: IMO we can add these (directly required) files under the `user` directory, although initially they're empty. This can prevent users from creating files in the wrong location, resulting in configuration not being adopted.

lua/core/event.lua
```diff
@@ -151,7 +151,10 @@ function autocmd.load_autocmds()
...
+	if ok then
+		definitions = vim.tbl_deep_extend("force", definitions, user_definitions)
+	end
```
Jint-lzxy, line 157:
```suggestion
	local ok, user_events = pcall(require, "user.event")
	if ok then
		definitions = vim.tbl_deep_extend("force", definitions, user_events)
	end
```
Better to follow naming conventions here.

lua/core/options.lua
```diff
@@ -133,3 +133,4 @@ local function load_options()
 end
 
 load_options()
+local _ = pcall(require, "user.options")
```
Jint-lzxy, line 136: Yet it is necessary to load user options:
```suggestion
local ok, user_options = pcall(require, "user.options")
if ok then
	for name, value in pairs(user_options) do
		vim.o[name] = value
	end
end
```

Those options can also be merged earlier on L124-126.

lua/core/pack.lua
```diff
@@ -5,6 +5,7 @@ local vim_path = global.vim_path
 local data_dir = global.data_dir
 local lazy_path = data_dir .. "lazy/lazy.nvim"
 local modules_dir = vim_path .. "/lua/modules"
+local user_modules_dir = vim_path .. "/lua/user/modules"
```
Jint-lzxy, line 8:
```suggestion
local user_modules_dir = vim_path .. "/lua/user"
```
It's okay to omit trailing `modules` here (e.g., `.../lua/user/plugins/*.lua` would be enough). The original intention of using `modules` is to follow `Neovim`'s guide on how to write plugins. This can also separate user configs from the default ones to avoid confusion.

Or you may exchange these two for minimum modifications:
```suggestion
local user_modules_dir = vim_path .. "/lua/modules/user"
```

lua/core/pack.lua
```diff
@@ -25,11 +26,16 @@ function Lazy:load_plugins()
...
 		package.path = package.path
 			.. string.format(";%s;%s", modules_dir .. "/configs/?.lua", modules_dir .. "/configs/?/init.lua")
+			.. string.format(";%s;%s", user_modules_dir .. "/configs/?.lua", user_modules_dir .. "/configs/?/init.lua")
 	end
 
 	local get_plugins_list = function()
 		local list = {}
 		local plugins_list = vim.split(fn.glob(modules_dir .. "/plugins/*.lua"), "\n")
+		local user_plugins_list = vim.split(fn.glob(user_modules_dir .. "/plugins/*.lua"), "\n")
+		if user_plugins_list[1] ~= "" then
```
Jint-lzxy, line 36:
```suggestion
		if not vim.tbl_isempty(user_plugins_list) then
```
Better to use native Lua API's to cope with possible corner cases.

lua/core/pack.lua
```diff
@@ -25,11 +26,16 @@ function Lazy:load_plugins()
...
+			plugins_list = vim.list_extend(plugins_list, user_plugins_list)
```
Jint-lzxy, line 37:
```suggestion
			vim.list_extend(plugins_list, user_plugins_list)
```
`list_extend` mutates `plugins_list`, so no need to assign the result to itself.

lua/core/pack.lua
```diff
@@ -25,11 +26,16 @@ function Lazy:load_plugins()
...
+		end
 		if type(plugins_list) == "table" then
```
Jint-lzxy, line 39:
```suggestion
```

IMO it's gd to remove this conditional statement, as `list_extend` always results in (list-like) tables.

lua/core/pack.lua
```diff
@@ -25,11 +26,16 @@ function Lazy:load_plugins()
...
 			for _, f in ipairs(plugins_list) do
```
Jint-lzxy, line 40: This loop also requires a change. This line:
```lua
list[#list + 1] = f:sub(#modules_dir - 6, -1)
```
Not always yields correct results, as the lengths of `modules_dir` and `user_modules_dir` are not the same. Better to add some sort of `string:find()` here.

lua/core/settings.lua
```diff
@@ -114,4 +114,8 @@ settings["dap_deps"] = {
 	"python", -- Python (debugpy)
 }
 
+local ok, user_settings = pcall(require, "user.settings")
+if ok then
+	settings = vim.tbl_deep_extend("force", settings, user_settings)
+end
```
Jint-lzxy, line 120: Since this pattern is used in many places, why not implement this as a function? E.g.:
```suggestion
require("...").config_extend(settings, "user.settings")
```

lua/keymap/init.lua
```diff
@@ -27,3 +27,9 @@ require("keymap.editor")
 require("keymap.lang")
 require("keymap.tool")
 require("keymap.ui")
+
+local ok, user_keymap = pcall(require, "user.keymap")
+if ok then
+	_G._user_map = true
+	require("modules.utils.keymap").amend(user_keymap, "_user_map")
+end
```
Jint-lzxy, line 35: IMO we can add `keymap.replace()` to the `keymap` module to avoid all these tweaks. See below for details.

EDIT: It seems that we need to modify LSP related mappings separately.

lua/modules/utils/init.lua
```diff
@@ -249,4 +249,38 @@ function M.tobool(value)
 	end
 end
 
+---@param plugin_name string @Plugin name
+---@param opts table @Default options for plugin
+---@param isVimPlugin? boolean @If plugin is made by vimscript
+---@param extraSetupFunc? function @When the plugin name that calls setup is different from `modules/configs/*.lua`
+function M.load_plugin(plugin_name, opts, isVimPlugin, extraSetupFunc)
```
Jint-lzxy, line 256:
```suggestion
---@param plugin_name string @Name of the plugin
---@param opts table @Default options for plugin
---@param vim_plugin? boolean @If this plugin is a Vimscript one
---@param setup_callback? function @Provide this if the plugin's setup function is not the usual one
function M.load_plugin(plugin_name, opts, vim_plugin, setup_callback)
```
Better to be consistent in naming here.

lua/modules/utils/init.lua
```diff
@@ -249,4 +249,38 @@ function M.tobool(value)
...
+	local ok, custom = pcall(require, "user.modules." .. plugin_name)
+	if ok then
+		local setupFunc = extraSetupFunc or require(plugin_name).setup
+		if custom == nil then
+			setupFunc(false)
+		elseif type(custom) == "table" then
+			assert(
+				not isVimPlugin,
+				"This plugin is not made by lua, so define options in functions (probably using `vim.g.*`)"
+			)
```
Jint-lzxy, line 267: u should avoid using `assert` or `error` in core utilities / during startup _(unless it's an exception that cannot be handled)_, as doing so will unwind the stack and block further function calls - which may lead to "dangling/unusable configs". Better to use `vim.notify` w/ `vim.log.levels.ERROR`, there're several examples above.

lua/modules/utils/init.lua
```diff
@@ -249,4 +249,38 @@ function M.tobool(value)
...
+			setupFunc(opts)
+		elseif type(custom) == "function" then
+			local user_opts = custom()
+			if type(user_opts) == "table" and not isVimPlugin then
+				setupFunc(user_opts)
+			end
+		else
+			error(
+				"Please return `nil` if you disable plugin or `table` if you override config or `function` if you replace config completely "
+			)
```
Jint-lzxy, line 278: Same as above.

lua/modules/utils/keymap.lua
```diff
@@ -101,12 +101,14 @@ local function amend(cond, mode, lhs, rhs, opts)
...
-		(map.desc and " / " .. map.desc or ""),
-	})
+	if cond ~= "" then
+		options.desc = table.concat({
+			"[" .. cond,
+			(options.desc and ": " .. options.desc or ""),
+			"]",
+			(map.desc and " / " .. map.desc or ""),
+		})
+	end
```
Jint-lzxy, line 111: This change might be unnecessary, see below for an overview.

lua/modules/utils/keymap.lua
```diff
@@ -128,12 +130,37 @@ local function modes_amend(cond, mode, lhs, rhs, opts)
...
+local function modes_unset(mode, lhs, buf)
+	if type(mode) == "table" then
+		for _, m in ipairs(mode) do
+			unset(m, lhs, buf)
+		end
+	else
+		unset(mode, lhs, buf)
+	end
+end
+
```
Jint-lzxy, line 156: I think a better implementation would be to \<Ctrl-CV\> `amend(cond, mode, lhs, rhs, opts)` and remove calls to generate the `fallback` function. `get_map()` will do its job (to del related keymaps).

Jint-lzxy (commented):
Several general comments:

Jint-lzxy (dismissed):
Some changes may require an update.

lua/core/event.lua
misumisumi: I created a new `lua/user` and created a plain file under it.
There is room for consideration for keymap.
Details below.

lua/core/event.lua
misumisumi: Resolved below

lua/core/settings.lua
misumisumi: Created a new `config_extend` function in `modules.utils`.
This resulted in a unified description.

lua/core/pack.lua
misumisumi: Fixed putting user defined plugins in `lua/user/{plugins,configs}`.

lua/modules/utils/keymap.lua
misumisumi: Created a new `replace` function in `modules.utils.keymap`.
I used `nvim_(buf_)set_keymap` because `vim.keymap.set` could not work well with buffered keymaps like lsp and gitsigns.
In order to make the file structure similar to below `keymap`, I included `lsp` in `user/keymap/completion.lus` and `gitsigns` configuration in `user/keymap/ui.lua`, but luacheck is commented out by default because it does not pass.
Do you have a better solution?

misumisumi:
@Jint-lzxy 
Thank you for your review.
Addressed each issue.
Details are in each comment.

lua/keymap/init.lua
misumisumi: Solved with the function suggested [below](https://github.com/ayamir/nvimdots/pull/931#discussion_r1287995599).

lua/modules/utils/init.lua
Jint-lzxy: > IMO, it might be more intuitive to create a `disabled_plugins` in settings.lua` and a `disabled_plugins` function in `modules.utils`.

+1. Im buying this lol

Jint-lzxy:
https://github.com/ayamir/nvimdots/pull/931/commits/f5ab67b6e0b7edeafb65733f09f507b67d3a82b8 did several comprehensive housekeeping and bug fixes, which includes the following changes:
- Added two recently modified plugins: `glance` and `symbols-outline`;
- Simplified some parts of the code and tried to follow naming conventions;
- Added several comments for code blocks with high cognitive complexity;
- Provided several "presets" (i.e., _examples_ for each user config file)

> In order to make the file structure similar to below `keymap`, I included `lsp` in `user/keymap/completion.lua` and `gitsigns` configuration in `user/keymap/ui.lua`, but luacheck is commented out by default because it does not pass.
> Do you have a better solution?

This is resolved by providing user default values (a.k.a., "presets"). I also deleted `gitsigns`-related mappings b/c users can directly override them in the corresponding config file :D

ayamir:
Seems `user` dir should be added to `.gitignore`.

misumisumi:
Two changes
- [f5ab67b](https://github.com/ayamir/nvimdots/pull/931/commits/f5ab67b6e0b7edeafb65733f09f507b67d3a82b8)
[f5ab67b](https://github.com/ayamir/nvimdots/pull/931/commits/f5ab67b6e0b7edeafb65733f09f507b67d3a82b8) Changes in no longer add `user/configs` to `package.path`. (I think it's a workaround for a potential bug with conflicting names)
This makes `config=require(user.configs.foo.bar)` very cumbersome if the user wants to use a structure similar to `lua/modules/plugins.lua`.
To solve this I added a custom loader as the lowest priority loader.
This is a loader that searches `user_configs_dir/configs` if `require(foo.bar)` does not exist in package.path.
-  [d73dc25](https://github.com/ayamir/nvimdots/pull/931/commits/d73dc251bfad295326cbfacc55b83392d10468f4)
Added `disabled_plugins` to settings based on [this review](https://github.com/ayamir/nvimdots/pull/931#discussion_r1288591275).
This allows users to disable plugins by simply writing the name of the plugin.

misumisumi:
@Jint-lzxy 
> A nasty implementation is used here: This snippet attempts to obtain the name of the file containing <upvalue> (debug.getinfo(2, ...)) and obtain the corresponding module name of the user config through a series of grepping (since we propose using the same file name as the original configuration to overwrite base config)- This might have some potential side effects I haven't figured out yet, so it requires some testing.

This implementation is a bit problematic.
`lsp.lua` sets up two plugins, `mason` and `mason-lspconfig`, but only references `user/completion/lsp.lua`.
This also applies to `null-ls.lua`.

Jint-lzxy:
> * [3583e26](https://github.com/ayamir/nvimdots/pull/931/commits/3583e26163aeb623bd7b4a1d41a36ca8653218f0)
>   [f5ab67b](https://github.com/ayamir/nvimdots/pull/931/commits/f5ab67b6e0b7edeafb65733f09f507b67d3a82b8) Changes in no longer add `user/configs` to `package.path`. (I think it's a workaround for a potential bug with conflicting names)

Yeah that's exactly the reason.

>   This makes `config=require(user.configs.foo.bar)` very cumbersome if the user wants to use a structure similar to `lua/modules/plugins.lua`.
>   To solve this I added a custom loader as the lowest priority loader.
>   This is a loader that searches `user_configs_dir/configs` if `require(foo.bar)` does not exist in package.path.

Maybe we can address this from another perspective? That is, only append `user_config_dir` to `package.path` when really needed (e.g., inside `M.load_plugin`). Since using bare `io` utils in the main event loop conflicts with [libuv](https://github.com/libuv/libuv)'s design (i.e., _asynchronous_ I/O). It is also feasible to directly use luv's API tho, but this is uniquely exposed to Neovim changes, both breaking and non-breaking.

> * [5722cf1](https://github.com/ayamir/nvimdots/pull/931/commits/5722cf102b9c65aebe05069b59ea189a7c82f148)
>   Support override lsp setting.
>   Use `tbl_deep_extend` to replace the corresponding value to make it easier to understand what the user is doing.

IMO we can leave those directories (lsp `servers`, dap `clients`, null-ls `sources`) as-is (or place them under the `user` directory) cause they're not subject to change. The thing is, if there _are_ breaking changes, users _must_ fix their config regardless of what we offer upstream - so introducing merge conflicts here sounds reasonable for me.

> IMO, it's easy to add an argument to bypass just using the given plugin name, but is there any other way?
> or divide the file

Divide them into separate components makes sense to me :)

Jint-lzxy:
Accidentally closed by hitting wrong button. Re-opening. Sorry.

misumisumi:
> Maybe we can address this from another perspective? That is, only append user_config_dir to package.path when really needed (e.g., inside M.load_plugin). Since using bare io utils in the main event loop conflicts with [libuv](https://github.com/libuv/libuv)'s design (i.e., asynchronous I/O). It is also feasible to directly use luv's API tho, but this is uniquely exposed to Neovim changes, both breaking and non-breaking.

The easiest way is to create a wrapper around `require` and let it resolve the path.
For example
```lua
M.load_config = function(modname)
    return require("user.configs." .. modname)
end
```
> Divide them into separate components makes sense to me :)

Work to divide.

> IMO we can leave those directories (lsp servers, dap clients, null-ls sources) as-is (or place them under the user directory) cause they're not subject to change. The thing is, if there are breaking changes, users must fix their config regardless of what we offer upstream - so introducing merge conflicts here sounds reasonable for me.

IMO, user custom settings and default settings should be completely separated.
To make it easier to debug with starting neovim without `user` if neovim doesn't start due to breaking changes as pointed out.
I think the most recent big change is probably the successor to `null-ls`, but I think you'll notice it because it won't be reflected even if the user expands the settings.
After that, I would encourage you to look at the release notes.

Jint-lzxy:
> This is a minor issue considering the impact on others.
> Let's write `user.configs` explicitly.

Ah I just realized that the `user` dir's hierarchy doesn't necessary have to follow our base config - this would make customization even more troublesome for users. So from this point of view, I think this is probably not too much typing for users? E.g.:
```lua
config = require("completion.cmp")
------
config = require("user.configs.cmp")
```
We can also take a step further - to include `lua/user` in our rtp. This way users only have to type:
```lua
config = require("configs.cmp")
```

> IMO, user custom settings and default settings should be completely separated. To make it easier to debug with starting neovim without `user` if neovim doesn't start due to breaking changes.

Agreed. Then perhaps we can move them directly to the `user` subdirectory - i.e., as presets? Since "dispersing" server configs can bring a lot of problems, especially when recursively merging server specs.

misumisumi:
> Ah I just realized that the user dir's hierarchy doesn't necessary have to follow our base config - this would make customization even more troublesome for users. So from this point of view, I think this is probably not too much typing for users? 

Yes, that's right.
Overrides for default installed plugins only need to be included in `lua/user/configs`, you don't need to imitate the structure if you add your own.

> We can also take a step further - to include lua/user in our rtp. This way users only have to type:

From what I've just checked, including `lua/user` in rtp seems like the best way to go.

> Then perhaps we can move them directly to the user subdirectory - i.e., as presets? Since "dispersing" server configs can bring a lot of problems, especially when recursively merging server specs.

You mean `cpmpletion/servers` and `dap/clients`, right?
As you said before, I think it's fine to include it in `user` if it allows for merge conflicts.
Because `function mason_lsp_handler` will start with default settings if there is no LSP setting. (neovim will start even if there is no setting for `user`.)
Or keep the current structure (you might create a new directory called `lua/presets/{servers,dap,...}`), and if you want to change the settings, you copy `lua/modules/configs/servers` to `lua/user/servers` or `lua/user/configs/servers`
  or `lua/user/configs/completion/servers` and edit it.
After that, `lsp.lua` will read `lua/user` first. (Do not merge.)
In other words, it's a way to use it as a preset and a fallback.

Jint-lzxy:
> In other words, it's a way to use it as a preset and a fallback.

Yeah I also agree that we can completely separate user custom settings and default settings. As you mentioned (if I understood this correctly), we should only provide two options: _overwrite_ or just _fallback_, as merging specs here can be a source of trouble. This also makes it more difficult for us to debug possible issues that may arise.

misumisumi:
If there are settings such as LSP under `lua/user`, read them, otherwise use presets.

> we might create a new directory called `lua/presets/{servers,dap,...}`

How about keeping the current structure or creating `lua/presets` to aggregate?

Jint-lzxy:
> How about keeping the current structure or creating `lua/presets` to aggregate?

IMO keeping current structure is more than enough. Because we also need to consider backward compatibility - some users may have already added their configs to the original directory (`completion/servers/*`...).

misumisumi:
Added the following changes:
- Separated `mason-lspconfig` and `mason` configuration from `lsp.lua`. (`mason-lspconfig.lua` and `mason.lua`, respectively)
- Separated `mason-null-ls` from `null-ls`. (`mason-null-ls.lua`)
- For `mason_{lsp,dap}_handler` if `lua/user/{lsp-servers,dap-clients}t` exists, prefer user settings
Assuming that the user does not match the structure under `user/configs` with the structure under `modules/configs`, the directory under `user` is referenced.
I added `lsp-` and `dap-` to the prefix for clarity.
- Add `lua/user` to rtp
- Added a conditional branch that just removes the keymap if the `[<modes>|<keymap>]="" or false` syntax is given

CharlesChiuGit (changes requested):
extraordinary refactor!
TBH, it's such a PITA to use custom settings before lol

lua/core/pack.lua
```diff
@@ -24,26 +25,30 @@ function Lazy:load_plugins()
...
 	local get_plugins_list = function()
 		local list = {}
 		local plugins_list = vim.split(fn.glob(modules_dir .. "/plugins/*.lua"), "\n")
-		if type(plugins_list) == "table" then
-			for _, f in ipairs(plugins_list) do
-				-- fill list with `plugins/*.lua`'s path used for later `require` like this:
-				-- list[#list + 1] = "plugins/completion.lua"
-				list[#list + 1] = f:sub(#modules_dir - 6, -1)
-			end
+		local user_plugins_list = vim.split(fn.glob(user_config_dir .. "/plugins/*.lua"), "\n", { trimempty = true })
```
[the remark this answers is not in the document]
Jint-lzxy: Nope b/c `split()` always returns a table (a.k.a, a vim List) :D
```console
split({string} [, {pattern} [, {keepempty}]])			*split()*
		Make a |List| out of {string}.  When {pattern} is omitted or
		...
```

CharlesChiuGit (approved):
LGTM

Jint-lzxy:
Maybe we need to rewrite [Wiki:Usage](https://github.com/ayamir/nvimdots/wiki/Usage) before merging this? A new release is also necessary IMO.

misumisumi:
Yeah, i need it.
Because the method of user custom has changed greatly.
Also, setting up an LSP server is different from setting up a plugin, so I think I have to explain.

CharlesChiuGit:
also i think it would nice to add a custom/user debugging issue template, so we can debug user's settings apart from our original one.

ayamir (approved):

CharlesChiuGit:
@misumisumi please check if there's any misunderstanding in the new issue template.
@Jint-lzxy there seem to have a syntax error in the new template, but I got no time to debug it right now. could u help me debug it?

[e93e52c](https://github.com/ayamir/nvimdots/pull/931/commits/e93e52c383c7fa806c4060e269b3f466868f5337)
[cd3456a](https://github.com/ayamir/nvimdots/pull/931/commits/cd3456aa4a0053e8e2e4477ddedcb5693efb3e16)
[45cebdf](https://github.com/ayamir/nvimdots/pull/931/commits/45cebdf33972eda6a4887ec47c042a1a411228de)
[74e79cf](https://github.com/ayamir/nvimdots/pull/931/commits/74e79cf69612cea1408ac19f9560978046488903)

CharlesChiuGit:
@misumisumi please check if there's any misunderstanding in the new issue template.

[0ec555d](https://github.com/ayamir/nvimdots/pull/931/commits/0ec555d64e96e292052faa81b1af91a57517c421)

Jint-lzxy:
@CharlesChiuGit I fixed several trivial formatting issues and added some comments. Now the `folder structure` section aligns pretty well :)

fs-ss

misumisumi:
LGTM!

misumisumi:
@Jint-lzxy 
I would appreciate it if you could reply to my review about `load_plugins`.

misumisumi:
@Jint-lzxy @CharlesChiuGit 
I think that `bug_report.yml` needs a check item to check if it has been started without `user`.
Also, if `user` is the cause, it is necessary to direct to `custom-config.yml`.

Jint-lzxy:
> I would appreciate it if you could reply to my review about `load_plugins`.

Oh sorry I missed that review, there are too many replies flying around this thread. May I ask which review you're referring to?

misumisumi:
Please review the `lua/modules/utils/init.lua` just above.

Jint-lzxy:
> I think that `bug_report.yml` needs a check item to check if it has been started without `user`.
> Also, if `user` is the cause, it is necessary to direct to `custom-config.yml`.

https://github.com/ayamir/nvimdots/pull/931/commits/72b99e12e0e2a9ef1ef6711e7836b49694a252b4

Jint-lzxy:
> Please review the `lua/modules/utils/init.lua` just above.

This looks pretty good to me lol

misumisumi:
> This looks pretty good to me lol

Do you mean that it's better to delete it?

Jint-lzxy:
> Do you mean that it's better to delete it?

I reviewed [this version](https://github.com/ayamir/nvimdots/pull/931/files#diff-76906787d569e73d36441b2ae9822c8617271189b4f24811d9d272e0e621cf0b) of `utils/init.lua` (which is marked as `changed since your last review`) and don't think we need to change anything. Have u pushed ur changes to GitHub?

misumisumi:
No, I didn't change anything.
I got it.
thank you.

CharlesChiuGit:
> @CharlesChiuGit I fixed several trivial formatting issues and added some comments. Now the `folder structure` section aligns pretty well :)

LGTM!

CharlesChiuGit:
Are we still waiting for the new wiki to complete?

Jint-lzxy:
> Are we still waiting for the new wiki to complete?

Yeah we're still waiting for that

misumisumi:
@Jint-lzxy @CharlesChiuGit 
Started a discussion about updating the wiki.

ayamir:
Can we merge it now?

misumisumi:
The code has been reviewed and the wiki has been edited, so it should be fine.

PR #964, opened 2023-08-18
fix: make rust-analyzer could auto start

ClSlaid:
patch refers to:
https://github.com/simrat39/rust-tools.nvim/issues/397#issuecomment-1609662279

### DIFF
--- a/lua/modules/plugins/lang.lua
+++ b/lua/modules/plugins/lang.lua
@@ -8,7 +8,6 @@ lang["fatih/vim-go"] = {
 }
 lang["simrat39/rust-tools.nvim"] = {
 	lazy = true,
-	ft = { "rust", "rs" },
 	event = { "BufReadPost *.rs" },
 	config = require("lang.rust-tools"),
 	dependencies = { "nvim-lua/plenary.nvim" },


ayamir:
Does this issue really exist? My `rust-analyzer` can start automatically when I `nvim main.rs`.

ClSlaid:
> Does this issue really exist? My `rust-analyzer` can start automatically when I `nvim main.rs`.

`rust-analyzer` that unmanaged by `mason-lspconfig` will not automatically start and attach.

Jint-lzxy:
I'm a bit dubious about whether this PR really fixes the problem. AFAIU it's possible that the `FileType` event is triggered _after_ the server(s) get attached, so the (newly) registered server isn't spawned by lspconfig automatically. Maybe a better solution is to explicitly tell lspconfig to load the new schema after `rust-tools` completes its startup (just blindly guessing, haven't verified this yet).

ayamir:
> I'm a bit dubious about whether this PR really fixes the problem.

No, it doesn't work for me though the `rust-analyzer` installed from `rustup` and it can be found in my `PATH`. Close it.

PR #985, opened 2023-09-05
fix: untrack changes in lua/user/ after installation.

ayamir:
Add a step after installation to make git `skip-worktree` from `user` dir.
For users already using, execute the command manually:
Windows:
```powershell
cd ~/AppData/Local/nvim
@(git ls-files lua/user) | ForEach-Object {&"git" update-index --skip-worktree $_}
```
*nix:
```shell
cd ~/.config/nvim
git ls-files -z lua/user/ | xargs -0 git update-index --skip-worktree
```
More info about `assume-unchanged` and `skip-worktree`: https://stackoverflow.com/questions/13630849/git-difference-between-assume-unchanged-and-skip-worktree

@popjdh

### DIFF
--- a/scripts/install.ps1
+++ b/scripts/install.ps1
@@ -287,6 +287,12 @@ function ring_bell {
 	[System.Console]::beep()
 }
 
+function cleanup {
+	info -Msg "Untracking user configs from git..."
+	safe_execute -WithCmd { Set-Location -Path "$env:CCDEST_DIR" }
+	@(git ls-files lua/user) | ForEach-Object {&"git" update-index --skip-worktree $_}
+}
+
 function _main {
 	if (-not $IsWindows) {
 		_abort -Msg "This install script can only execute on Windows." -Type "DeviceError"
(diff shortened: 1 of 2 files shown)


CharlesChiuGit (approved):
Can confirm it works on macos.

popjdh:
Just now, I changed `lua/user` to `lua/user_template` through `git mv`. And also modified the installation script to automatically generate `lua/user`.

This is the [PR #986](https://github.com/ayamir/nvimdots/pull/986) I submitted after modification.

ayamir:
@popjdh I prefer this way to do the same thing instead of introducing template files due to redundancy.
Besides, using command to ignore won't influence current users who already customized their configs.

popjdh:
> @popjdh I prefer this way to do the same thing instead of introducing template files due to redundancy. Besides, using command to ignore won't influence current users who already customized their configs.

1. For some users who use local branch to manage configurations, `git --skip-worktree` will disrupt their management
2. When modifying the preset `lua/user`, it is necessary to first `git --no-skip-worktree` before merging conflicts. Unless the preset is never changed, the consequences are equally cumbersome.

popjdh:
> @popjdh I prefer this way to do the same thing instead of introducing template files due to redundancy. Besides, using command to ignore won't influence current users who already customized their configs.

In addition, for users who have already updated to the latest version, it is not a difficult problem to backup using methods such as "cp", "tar", etc😘

ayamir:
> > @popjdh I prefer this way to do the same thing instead of introducing template files due to redundancy. Besides, using command to ignore won't influence current users who already customized their configs.
> 
>     1. For some users who use local branch to manage configurations, `git --skip-worktree` will disrupt their management
> 
>     2. When modifying the preset `lua/user`, it is necessary to first `git --no-skip-worktree` before merging conflicts. Unless the preset is never changed, the consequences are equally cumbersome.

You are correct, close it.

PR #993, opened 2023-09-10
feat(plugins)!: remove some preset colorschemes

ayamir:
1. In fact, `catppuccin` is enough for `nvimdots`.
2. Users could install other colorschemes by themselves.
3. There are several different versions of `nord` theme, users have the right to install which of them.


Jint-lzxy (approved):
Makes sense to me :D

CharlesChiuGit (approved):
lgtm!

PR #996, opened 2023-09-13
feat(null_ls_deps): add `gofumpt` and `gopls`

cherrot:
- Enable `gofumpt` in `gopls` setting since it's widely used in Go community
- Import missing packages via `goimports` on save.

### DIFF
--- a/lua/core/settings.lua
+++ b/lua/core/settings.lua
@@ -111,6 +111,8 @@ settings["null_ls_deps"] = {
 	"shfmt",
 	"stylua",
 	"vint",
+	"goimports",
+	"gofumpt",
 }
 
 -- Set the Debug Adapter Protocol (DAP) clients that will be installed and configured during bootstrap here.
--- a/lua/core/settings.lua
+++ b/lua/core/settings.lua
@@ -107,12 +107,12 @@ settings["lsp_deps"] = {
 ---@type string[]
 settings["null_ls_deps"] = {
 	"clang_format",
+	"gofumpt",
+	"goimports",
 	"prettier",
 	"shfmt",
 	"stylua",
 	"vint",
-	"goimports",
-	"gofumpt",
 }
 
 -- Set the Debug Adapter Protocol (DAP) clients that will be installed and configured during bootstrap here.


lua/modules/configs/completion/null-ls.lua
```diff
@@ -25,6 +25,7 @@ return function()
 			},
 		}),
 		btns.formatting.rustfmt,
+		btns.formatting.goimports,
```
Jint-lzxy, line 28: Better to move this to `settings["null_ls_deps"]` for automated installs & setups. This table is for sources that are either not registered in `mason-registry` or require some sort of customization.

Jint-lzxy (changes requested):

lua/modules/configs/completion/null-ls.lua
cherrot: I also added `gofumpt` to `settings["null_ls_deps"]`  (which is also registered in [null-ls' builtin](https://github.com/jose-elias-alvarez/null-ls.nvim/blob/main/lua/null-ls/builtins/formatting/gofumpt.lua)). 

Could you teach me how to test the automated install feature?  Thanks in advance :)

cherrot: Cool!

lua/modules/configs/completion/servers/gopls.lua
```diff
@@ -5,6 +5,7 @@ return {
 	settings = {
 		gopls = {
 			usePlaceholders = true,
+			gofumpt = true,
```
Jint-lzxy, line 8: Is it still necessary to set `settings.gofumpt` here? We've already registered that as a null-ls dep (source).

cherrot: You are right. I'd better to leave the minimal change :)

Jint-lzxy (approved):
LGTM 🍯

ayamir (approved):

Saafo:
How to disable these two in user setting? Merging will not remove.

CharlesChiuGit:
@Saafo take another look at https://github.com/ayamir/nvimdots/wiki/Usage#about-user-overrides-settings-options-events-etc

PR #1003, opened 2023-09-20
fix(neodim): pin to v2.0

Jint-lzxy:
Neodim's recent updates (specifically https://github.com/zbirenbaum/neodim/commit/2ac2f25d221bf34f70b79d0ab1d313e9826859c3) introduced/utilized some of `nvim 0.10`'s APIs, which might be a source of trouble for nvim 0.9.\*, so IMO we'd better pin `neodim` to v2.0 to avoid potential issues :D


CharlesChiuGit (approved):
agreed

ayamir (approved):

PR #1016, opened 2023-10-10
fix(nixos): Add go lang.

misumisumi:
Go-related lang support was added in #996.
When launching neovim on NixOS, building these tools fails due to missing `go`.
Therefore, add `go` to `programs.neovim.extraPackages`.


ayamir (approved):
Make sense but I can't test it.

CharlesChiuGit (commented):
same as ayamir lol, but i think we can merge it first.

Issue #970, opened 2023-08-24
C-k <bs> not work in terminal windows


gxing:
### Version confirmation

- [X] Confirm

### Following prerequisites

- [X] Confirm

### Neovim version

0.9.1

### Operating system/version

Linux

### Terminal name/version

iTerm2

### $TERM environment variable

xterm-256color

### Branch info

main (Default/Latest)

### Fetch Preferences

HTTPS (use_ssh = false)

### How to reproduce the issue

use 
:Terminal 
or A-d 
or C-\ 
or A-\
open terminal 

C-k and  backupspace

### Expected behavior

delete to end of line(C-k)
delete previous  char ( backspace)

### Actual behavior

terminal loss focus

### Additional information

found in 

lua/core/mapping.lua (line 20-23)

	["t|<C-h>"] = map_cmd("<Cmd>wincmd h<CR>"):with_silent():with_noremap():with_desc("window: Focus left"),
	["t|<C-l>"] = map_cmd("<Cmd>wincmd l<CR>"):with_silent():with_noremap():with_desc("window: Focus right"),
	["t|<C-j>"] = map_cmd("<Cmd>wincmd j<CR>"):with_silent():with_noremap():with_desc("window: Focus down"),
	["t|<C-k>"] = map_cmd("<Cmd>wincmd k<CR>"):with_silent():with_noremap():with_desc("window: Focus up"),

ayamir:
We defined `Ctrl` and `h`, `j`, `k`, and `l` to focus switch at `lua/core/mapping.lua`.
A common requirement is when you use `C-\\` or `A-\\` to open the horizontal or vertical terminal, you want to use the keyboard shortcut to switch to the code window and keep the terminal not hidden.
If you need the native bash shortcut, you can remove these lines.

ayamir:
Close due to inactivity and being anwsered.

PR #1029, opened 2023-10-19
fix(keymap): use `<C-w>` with hjkl to navigate under `t` mode.

ayamir:
As discussed in #970, the keymaps of `<C-h>`, `<C-j>`, `<C-k>`, and `<C-l>` are usually defined as shell keymaps to do other things. For example, `<C-l>` is used to clear terminal output, which is used frequently. Old mappings just considered the consistency with those under normal mode and didn't consider the conflicts.


CharlesChiuGit (approved):
agreed, LGTM

Jint-lzxy (approved):

PR #1031, opened 2023-10-20
fix(lualine): automatically load theme for the active colorscheme

ayamir:
By utilizing "auto" theme as mentioned [here](https://github.com/nvim-lualine/lualine.nvim/blob/master/THEMES.md#auto), this PR corrects the appearance of `lualine` when using other colorschemes.
Of course, users using `catppuccin` won't be influenced.

### DIFF
--- a/lua/modules/configs/ui/lualine.lua
+++ b/lua/modules/configs/ui/lualine.lua
@@ -1,6 +1,5 @@
 return function()
-	local colorscheme = vim.g.colors_name
-	local _is_catppuccin = string.find(colorscheme, "catppuccin") ~= nil
+	local is_catppuccin = vim.g.colors_name:find("catppuccin") ~= nil
 	local colors = require("modules.utils").get_palette()
 	local icons = {
 		diagnostics = require("modules.utils.icons").get("diagnostics", true),
@@ -10,14 +9,13 @@ return function()
 		ui = require("modules.utils.icons").get("ui", true),
 	}
 
-	local function custom_theme(is_catppuccin)
+	local function custom_theme()
+		is_catppuccin = vim.g.colors_name:find("catppuccin") ~= nil
 		vim.api.nvim_create_autocmd("ColorScheme", {
 			group = vim.api.nvim_create_augroup("LualineColorScheme", { clear = true }),
 			pattern = "*",
 			callback = function()
-				local colorscheme_ = vim.g.colors_name
-				local is_catppuccin_ = string.find(colorscheme_, "catppuccin") ~= nil
-				require("lualine").setup({ options = { theme = custom_theme(is_catppuccin_) } })
+				require("lualine").setup({ options = { theme = custom_theme() } })
 			end,
 		})
 
@@ -144,7 +142,7 @@ return function()
 				return "│"
 			end,
 			padding = 0,
-			color = _is_catppuccin and utils.gen_hl("surface1", true, true) or nil,
+			color = is_catppuccin and utils.gen_hl("surface1", true, true) or nil,
 		},
 
 		file_status = {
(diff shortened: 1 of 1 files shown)


Jint-lzxy (commented):
I mean we can keep changes minimal [like this](https://github.com/Jint-lzxy/nvimdots/commit/d1ded4d3c1daf8341517c38015b62658defed087) b/c our palette will also "try it's best to generate one if there's no theme available for the currently selected colorscheme."

ayamir:
图片
IMO it's better to `return "auto"` here b/c our code based on `catppuccin` may not suitable for every theme like `gruvbox` in the screenshot.

CharlesChiuGit (approved):
TBH, I cant really tell the difference between two proposes in my own tests.
But ayamir's propose seems to use less code without changing the visual effect, so I think this PR is fine.

CharlesChiuGit (approved):
TBH, I cant really tell the difference between two proposes in my own tests.
But ayamir's propose seems to use less code without changing the visual effect, so I think this PR is fine.

CharlesChiuGit (approved):
TBH, I cant really tell the difference between two proposes in my own tests.
But ayamir's propose seems to use less code without changing the visual effect, so I think this PR is fine.

Jint-lzxy:
> IMO it's better to `return "auto"` here b/c our code based on `catppuccin` may not suitable for every theme like `gruvbox` in the screenshot.

Agree with this lol I was just wondering if lualine's [implementation of the `auto` theme](https://github.com/nvim-lualine/lualine.nvim/blob/2248ef254d0a1488a72041cfb45ca9caada6d994/lua/lualine/themes/auto.lua) could lead to performance overheads b/c we basically only need the function to check whether there's a theme available for the current colorscheme (and if there is, load that instead of generating ours). So I only took the relevant parts from `lualine/themes/auto.lua` [in this proposal](https://github.com/Jint-lzxy/nvimdots/commit/d1ded4d3c1daf8341517c38015b62658defed087) :D

I also think maybe we can simplify our implementation a bit lol (cause the modifications to those `color` keys look "quite" similar). Would it be better to just use a status variable to indicate the currently active theme and tweak `utils.gen_hl` to generate the colors where appropriate?

cc @CharlesChiuGit @ayamir

CharlesChiuGit:
Aha, I see your point now. ys, I think it's better to tweak `utils.gen_hl` to generate colors would be a nice approach.

ayamir:
It seems that the performance overhead is insignificant after I tested the two implementations using `vim-startuptime`.
image
image

lua/modules/configs/ui/lualine.lua
```diff
@@ -8,46 +10,6 @@ return function()
...
-			replace = {
-				a = { fg = colors.red, bg = colors.surface0, gui = "bold" },
-			},
-			inactive = {
-				a = { fg = colors.subtext0, bg = universal_bg, gui = "bold" },
-				b = { fg = colors.subtext0, bg = universal_bg },
-				c = { fg = colors.subtext0, bg = universal_bg },
-			},
-		}
-	end
```
Jint-lzxy, line 49: I think we still need this function b/c it registered an autocmd that would automatically reload lualine when switching colorscheme(s) and also defined colors that look more "concinnous" for this configuration (the colors of our components are based on this, and the (preset) style (catppuccin-\*) actually differs slightly from ours.) FWIW this will require some sort of adjustments if we plan to use the `auto` theme (changes mainly to modes other than `normal`)

lua/modules/configs/ui/lualine.lua
```diff
@@ -137,7 +99,7 @@ return function()
 				return "│"
 			end,
 			padding = 0,
-			color = utils.gen_hl("surface1", true, true),
+			color = is_catppuccin and utils.gen_hl("surface1", true, true) or nil,
```
Jint-lzxy, line 102: IMO we can use a status variable to indicate the currently active theme and tweak `utils.gen_hl` to generate the colors where appropriate instead. (https://github.com/ayamir/nvimdots/pull/1031#issuecomment-1774631337)

ayamir:
Using "auto" for `catppuccin` differs from our customized appearance. So I reverted to the initial commit status and unified code styles. It satisfied the requirement as I mentioned [here](https://github.com/ayamir/nvimdots/pull/1031#issue-1954382100).

Jint-lzxy (approved):
LGTM!

PR #1033, opened 2023-10-21
feat: customized `neovide` and `gui` config

ayamir:
This PR did two things.

+ Separate the config of `guifont` and `neovide`.
+ Add two settings entrys in `lua/core/settings.lua` named as `gui_config` and `neovide_config`.

This PR is backward-compatible.


CharlesChiuGit (approved):
LGTM

PR #1039, opened 2023-10-25
fix: check `type(value)==table` before recursion.

ayamir:
https://github.com/ayamir/nvimdots/discussions/806#discussioncomment-7378476


Jint-lzxy (approved):
LGTM

PR #1052, opened 2023-11-03
fix(neovide): neovide is not set up correctly because of `ipairs`

csyJoy:
since the numeric keys are ignored as whole in `iparis` iteration, configurations in `neovide_config` actually aren't set to neovide, after switching `ipairs` to `pairs`, configurations are set properly.


ayamir (approved):
My fault.

Issue #1058, opened 2023-11-09
Symbol outline doesn't work properly in python files and C files


Qejun:
### Version confirmation

- [X] Confirm

### Following prerequisites

- [X] Confirm

### Not a user config issue

- [X] Confirm

### Neovim version

nvim0.9.4

### Operating system/version

win10

### Terminal name/version

Windows terminal 

### Branch info

main (Default/Latest)

### Fetch Preferences

SSH (use_ssh = true)

### How to reproduce the issue

Symbol outline doesn't work properly in python files and C files，I wonder where to start with this problem
![屏幕截图 2023-11-09 080019](https://github.com/ayamir/nvimdots/assets/24759432/b9121a93-2b6b-4c32-89b9-4ceb04fda7f2)

### Expected behavior

In a lua file, the symbol outline correctly displays the hierarchical structure of the content

![屏幕截图 2023-11-09 080714](https://github.com/ayamir/nvimdots/assets/24759432/b501a452-188c-484f-98b0-90851213cdf0)

### Actual behavior

In python files and C files, everything is displayed on the same level when navigating with symbol outline
![屏幕截图 2023-11-09 080019](https://github.com/ayamir/nvimdots/assets/24759432/68b13266-7ba3-4715-bd5e-adf531ae40e7)

Qejun:
<img width="960" alt="屏幕截图 2023-11-09 093816" src="https://github.com/ayamir/nvimdots/assets/24759432/5c3262aa-0605-4d4c-bc8e-51ff8b91c2db">
I changed python's lsp and now it works correctly.
Change python-lsp-server to pyright in lua/modules/configs/completion/mason.lua
However, there is still no way to solve the c file.Do you have any recommended c language server?

CharlesChiuGit:
ys, the syntax tree is generate from treesitter and lsp. glad to see u try out different lsp on python.

for c/c++ part, what c lsp r u using now? what c lsp have u tried?

Jint-lzxy:
Hmm, I think this is more of an "upstream support" issue. Below are some results of trying to communicate with `pylsp` and `clangd` via json-rpc (correct me if im wrong 😃):
- Consider the following code for `pylsp` (taken from [here](https://github.com/intel/gvt-linux/blob/e0598d0ee7f9eba4cbee832e11468e2d816d1051/tools/net/ynl/lib/ynl.py#L70-L97)):
```python
class NlAttr:
    def __init__(self, raw, offset):
[...]

PR #1059, opened 2023-11-09
refactor: use aerial to provide outline.

ayamir:
[Aerial.nvim](https://github.com/stevearc/aerial.nvim) has better outline support than [symbols-outline.nvim](https://github.com/simrat39/symbols-outline.nvim). #1058 will be solved.


ayamir:
image

lua/modules/configs/completion/aerial.lua
```diff
@@ -0,0 +1,352 @@
...
+		-- To see all available values, see :help SymbolKind
+		filter_kind = {
+			"Class",
+			"Constructor",
+			"Enum",
+			"Function",
+			"Interface",
+			"Module",
+			"Method",
+			"Struct",
```
CharlesChiuGit, line 110: I think it's too few symbols for default usage, maybe add all symbols first?
```
Array
Boolean
Class
Constant
Constructor
Enum
EnumMember
Event
Field
File
Function
Interface
Key
Method
Module
Namespace
Null
Number
Object
Operator
Package
Property
String
Struct
TypeParameter
Variable
```

CharlesChiuGit (changes requested):

lua/modules/configs/completion/aerial.lua
CharlesChiuGit: before
image

after
image

CharlesChiuGit:
also the keymaps doesn't seem working.

CharlesChiuGit:
also i think `aerial`'s telescope extension is so spot on that I think we should just change `go` to open the telescope preview.
How do u think? @ayamir @Jint-lzxy

Jint-lzxy:
> also i think `aerial`'s telescope extension is so spot on that I think we should just change keymap:`go` to open it's telescope preview. How do u think?

Agree, but why not assign a keymap to each of them lol

CharlesChiuGit:
> Agree, but why not assign a keymap to each of them lol

absolutely! that would be nice too!

CharlesChiuGit (approved):

CharlesChiuGit:
I think it gd to go.

Jint-lzxy:
Sorry I'm a bit busy with other stuff atm. I'll review this PR sometime tomorrow :D

CharlesChiuGit (approved):
LGTM

Jint-lzxy (approved):
LGTM 🍯 but we probably need to update the relevant parts of the wiki before merging this lol

PR #1062, opened 2023-11-11
feat(treesitter): use ts-indent and disable legacy syntax engine

Jint-lzxy:
https://github.com/nvim-treesitter/nvim-treesitter/commit/f53e6e82df8efb17b9b04d2f6d5ad64755786b7e fixed the last dangling syntax issue in C/C++, So IMO we can now completely switch to treesitter-based highlight (and indent) lol


CharlesChiuGit (approved):
not a c expert, but the commit looks convincing.

ayamir (approved):

Issue #1077, opened 2023-11-24
"disabled_plugins" setting won't work for forked plugins


YuCao16:
### Version confirmation

- [X] Confirm

### Following prerequisites

- [X] Confirm

### Not a user config issue

- [X] Confirm

### Neovim version

NVIM v0.10.0-dev-618+g4fd852b8c

### Operating system/version

Ubuntu 23.10

### Terminal name/version

kitty 0.28.1

### $TERM environment variable

xterm-kitty

### Branch info

main (Default/Latest)

### Fetch Preferences

SSH (use_ssh = true)

### How to reproduce the issue

1. fork a plugin's repo, for example `nvimdev/lspsaga.nvim` as `username/lspsaga.nvim`, and add it to `user/plugins/completion.lua`, and in settings, add settings["disabled_plugins"] = {"nvimdev/lspsaga.nvim"}. 
2. Restart neovim and run `Lazy update`
3. Enter neovim and run command `Lspsaga xxx`
4. It will return not a known command

### Expected behavior

The `nvimdev/lspsaga.nvim` will be disabled, and I will use `username/lspsaga.nvim`. 

### Actual behavior

`Lspsaga xxx` is not a known command.

### Additional information

This is not a problem of nvimdots actually, it related to `lazy.nvim`, as nvimdots use disable option to disable the plugin. However, `lazy.nvim` seems not support such situation. Maybe remove those disabled plugin from here?
https://github.com/ayamir/nvimdots/blob/f84f72da4b0054fe8871c70431c5329a621dbb8f/lua/core/pack.lua#L40-L44

Let me known if you need more steps to reproduce!

ayamir:
The problem is `lazy.nvim` store all plugins in `~/.local/share/nvim/site/lazy` using just the plugin name (`lspsaga.nvim`) rather than with the author name (`nvimdev/lspsaga.nvim`). So it won't clone the forked repo because their names are the same.

Jint-lzxy:
[...]

PR #1079, opened 2023-11-26
fix: don't add the disabled plugins to plugins list.

ayamir:
#1077 
@Jint-lzxy @CharlesChiuGit @YuCao16

### DIFF
--- a/lua/core/pack.lua
+++ b/lua/core/pack.lua
@@ -48,17 +48,22 @@ function Lazy:load_plugins()
 
 	append_nativertp()
 
+	-- init hashtable for disabled_plugins
+	local disabled_plugins_table = {}
+	for _, name in ipairs(settings.disabled_plugins) do
+		disabled_plugins_table[name] = true
+	end
 	for _, m in ipairs(get_plugins_list()) do
 		-- require modules returned from `get_plugins_list()` function.
 		local modules = require(m:sub(0, #m - 4))
 		if type(modules) == "table" then
 			for name, conf in pairs(modules) do
-				self.modules[#self.modules + 1] = vim.tbl_extend("force", { name }, conf)
+				-- add plugin which not in disabled_plugins_table
+				if not disabled_plugins_table[name] then
+					self.modules[#self.modules + 1] = vim.tbl_extend("force", { name }, conf)
+				end
 			end
 		end
-	end
-	for _, name in ipairs(settings.disabled_plugins) do
-		self.modules[#self.modules + 1] = { name, enabled = false }
 	end
 end
 
--- a/lua/core/pack.lua
+++ b/lua/core/pack.lua
@@ -46,12 +46,36 @@ function Lazy:load_plugins()
 		return list
 	end
 
+	-- remove plugins in (dependencies and disabled_plugins_table) recursively
+	local function iterate_dependencies(dependencies, disabled_plugins_table)
+		for i = #dependencies, 1, -1 do
+			local dependency = dependencies[i]
+			if type(dependency) == "table" then
+				for key, value in pairs(dependency) do
+					if key == "dependencies" and type(value) == "table" then
+						iterate_dependencies(value, disabled_plugins_table)
+					elseif key == 1 and disabled_plugins_table[value] then
+						table.remove(dependencies, i)
+					end
+				end
+			elseif type(dependency) == "string" and disabled_plugins_table[dependency] then
+				table.remove(dependencies, i)
+			end
+		end
+	end
+
+	local remove_in_dependencies = function(conf, disabled_plugins_table)
+		if conf["dependencies"] ~= nil then
+			iterate_dependencies(conf["dependencies"], disabled_plugins_table)
+		end
+	end
+
 	append_nativertp()
 
 	-- init hashtable for disabled_plugins
 	local disabled_plugins_table = {}
-	for _, name in ipairs(settings.disabled_plugins) do
-		disabled_plugins_table[name] = true
+	for _, plugin in ipairs(settings.disabled_plugins) do
+		disabled_plugins_table[plugin] = true
 	end
 	for _, m in ipairs(get_plugins_list()) do
 		-- require modules returned from `get_plugins_list()` function.
@@ -60,6 +84,7 @@ function Lazy:load_plugins()
 			for name, conf in pairs(modules) do
 				-- add plugin which not in disabled_plugins_table
 				if not disabled_plugins_table[name] then
+					remove_in_dependencies(conf, disabled_plugins_table)
 					self.modules[#self.modules + 1] = vim.tbl_extend("force", { name }, conf)
 				end
 			end


aarnphm (approved):

YuCao16:
> #1077 @Jint-lzxy @CharlesChiuGit @YuCao16

It works, thanks. 

By the way, I noticed that all the user plugins' config are stored in `user/configs/*.lua`, but since I got too many additional plugins, the `user/configs/` directory got messy, can we introduce the structures just as the `/modules/configs/`, for example `user/configs/ui/*.lua`?

CharlesChiuGit (approved):
LGTM

YuCao16:
@ayamir @CharlesChiuGit 
I found a problem, this method seems not work for dependencies plugin, for example `nvim-treesitter/nvim-treesitter`, I cannot disable the dependencies such as `andymass/vim-matchup`.

Let's say I don't use `andymass/vim-matchup`, and I need another dependencies `anuvyklack/pretty-fold.nvim`. So in `user/plugins/editor`, I add following code (not sure this is the correct way?):
``` lua
	lazy = true,
	build = function()
		if #vim.api.nvim_list_uis() ~= 0 then
			vim.api.nvim_command([[TSUpdate]])
		end
	end,
	event = "BufReadPre",
	config = require("editor.treesitter"),
	dependencies = {
                {" anuvyklack/pretty-fold.nvim" },
		{ "andymass/vim-matchup" },
		{ "mfussenegger/nvim-treehopper" },
		{ "nvim-treesitter/nvim-treesitter-textobjects" },
		{
			"abecodes/tabout.nvim",
			config = require("editor.tabout"),
		},
		{
			"windwp/nvim-ts-autotag",
			config = require("editor.autotag"),
		},
		{
			"NvChad/nvim-colorizer.lua",
			config = require("editor.colorizer"),
		},
		{
			"hiphish/rainbow-delimiters.nvim",
			config = require("editor.rainbow_delims"),
		},
		{
			"nvim-treesitter/nvim-treesitter-context",
			config = require("editor.ts-context"),
		},
		{
			"JoosepAlviste/nvim-ts-context-commentstring",
			config = require("editor.ts-context-commentstring"),
		},
	},
}
```
and set `disabled_plugins = {"andymass/vim-matchup"}`, but `andymass/vim-matchup` still enabled as current solution don't modify the dependencies.

ayamir:
@YuCao16 Bug free now.

ayamir:
> By the way, I noticed that all the user plugins' config are stored in user/configs/*.lua, but since I got too many additional plugins, the user/configs/ directory got messy, can we introduce the structures just as the /modules/configs/, for example user/configs/ui/*.lua?

It can be done in another PR.

Jint-lzxy (commented):
Hmm I'm a bit opposed to this PR b/c it not only adds a layer of complexity, but also addresses only one potential "edge use case". This is why folke said in [this issue](https://github.com/folke/lazy.nvim/issues/861) that he won't support including two plugins with the same (plugin) name but different repo names in the spec at the same time. Instead, he suggested that ppl should use the `dev` (or `url`) option, e.g.:
```lua
completion["nvimdev/lspsaga.nvim"] = {
	url = "https://example.com/path/to/ur/lspsaga.nvim.git",
	-- other configs
}
```

Jint-lzxy:
> > By the way, I noticed that all the user plugins' config are stored in `user/configs/*.lua`, but since I got too many additional plugins, the `user/configs/` directory got messy, can we introduce the structures just as the `/modules/configs/`, for example `user/configs/ui/*.lua`?
>
> It can be done in another PR.

I think we already support this now:
```lua
editor["m4xshen/autoclose.nvim"] = {
	lazy = true,
	event = "InsertEnter",
	config = require("configs.editor.autoclose"), -- will require user/configs/editor/autoclose.lua
}
```
> Let's say I don't use `andymass/vim-matchup`, and I need another dependencies `anuvyklack/pretty-fold.nvim`. So in `user/plugins/editor`, I add following code (not sure this is the correct way?):

Because `lazy.nvim` will also (internally) merge the plugin spec it manages, you only need to write:
``` lua
	dependencies = {
                {" anuvyklack/pretty-fold.nvim" },
		{ "andymass/vim-matchup", enabled = false },
	},
}
```

Jint-lzxy:
> > Hmm I'm a bit opposed to this PR b/c it not only adds a layer of complexity, but also addresses only one potential "edge use case". This is why folke said in [this issue](https://github.com/folke/lazy.nvim/issues/861) that he won't support including two plugins with the same (plugin) name but different repo names in the spec at the same time. Instead, he suggested that ppl should use the `dev` (or `url`) option, e.g.:
> > ```lua
> > completion["nvimdev/lspsaga.nvim"] = {
> > 	url = "https://example.com/path/to/ur/lspsaga.nvim.git",
> > 	-- other configs
> > }
> > ```
> 
> But seems it requires users to change repo's config. And I think we should keep our API for users works correctly for all of use cases.

I just checked [the implementation](https://github.com/folke/lazy.nvim/blob/96584866b9c5e998cbae300594d0ccfd0c464627/lua/lazy/core/plugin.lua#L438-L471) of `lazy.nvim`, and it appears that when several plugins with the same name don't include conflicting URLs, `lazy` will use the source declared by _the last appended specification_ (i.e., the spec provided by the user). The following minimal example demonstrates this:
```lua
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
	vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
	vim.fn.system({
		"git",
		"clone",
		"--filter=blob:none",
		"--single-branch",
		"https://github.com/folke/lazy.nvim.git",
		lazypath,
	})
end
vim.opt.runtimepath:prepend(lazypath)

-- install plugins
local plugins = {
	"folke/tokyonight.nvim",
	-- add any other pugins here
	{ "nvimdev/lspsaga.nvim" },
	{ "ayamir/lspsaga.nvim", opts = {} },
}
require("lazy").setup(plugins, {
	root = root .. "/plugins",
})

vim.opt.termguicolors = true
vim.cmd([[colorscheme tokyonight]])
```
Save it as `init.lua`, and running it with `nvim -u init.lua` yields the following (possible) result:
```console
  ● lspsaga.nvim 12.42ms ▷ start
      dir    /private/var/folders/jy/91xg_8j9169dydct2k4r7yw80000gn/T/tmp.chC1vsz66n/.repro/plugins/lspsaga.nvim
      url    https://github.com/ayamir/lspsaga.nvim
                                ^^^^^^
      branch main
      commit d3dfaea
      readme README.md
      help   |lspsaga.nvim.txt|
```
So IMO our loader has _indeed_ done what it was _supposed_ to do. The only thing missing here is we need to document this particular use case in the wiki - if users want to use their own fork(s) to _overwrite_ the default one(s), they don’t need to include the base plugin(s) in `disabled_plugins`.

ayamir:
OK, IMO this pr and #1077 can be closed now. @YuCao16

PR #1081, opened 2023-11-26
feat: add docker support.

ayamir:
I built a docker image for `nvimdots`: https://hub.docker.com/repository/docker/ayamir/nvimdots/general

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,8 @@ RUN apt-get update && apt-get install -y \
 RUN update-ca-certificates
 
 # Add a command to select a mirror based on a condition
-ARG use_mirror=false
+ARG USE_MIRROR=false
+ENV use_mirror=$USE_MIRROR
 RUN if [ "$use_mirror" = "true" ]; then \
     sed -i 's@http://archive.ubuntu.com/ubuntu/@https://mirrors.tuna.tsinghua.edu.cn/ubuntu/@' /etc/apt/sources.list \
     ;fi
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,125 @@
+# Use Ubuntu 22.04 as the base image
+FROM ubuntu:22.04
+
+# Set environment variables
+ENV DEBIAN_FRONTEND=noninteractive
+ENV CMAKE_EXTRA_FLAGS=-DENABLE_JEMALLOC=OFF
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,125 @@
...
+
+# Install the ca-certificates package to update CA certificates
+RUN apt-get update && apt-get install -y \
+    ca-certificates
+
+# Update CA certificates
+RUN update-ca-certificates
+
+# Add a command to select a mirror based on a condition
+ARG use_mirror=false
```
aarnphm, line 16: more docker-ich

```suggestion
ARG USE_MIRROR=false
ENV use_mirror=$USE_MIRROR
```

Dockerfile
```diff
@@ -0,0 +1,125 @@
...
+RUN git clone https://github.com/neovim/unibilium.git && \
+    cd unibilium && \
+    autoreconf -fi && \
+    ./configure --prefix=/usr/local && \
+    make && \
+    make install && \
+    cd ../ && rm -rf unibilium
+
+# Install the latest stable neovim
+RUN curl -LO https://github.com/neovim/neovim/releases/download/stable/nvim-linux64.tar.gz \
```
aarnphm, line 84: should we include two variants? One stable and one nightly?

Probably can use multi-build here.

Dockerfile
```diff
@@ -0,0 +1,125 @@
...
+WORKDIR /root
+
+# Set TERM
+ENV TERM=xterm-256color
+
+# Bootstrap
+RUN nvim --headless "+Lazy! sync" +qa
+
+# Start with bash for the convenience of new changes
+CMD ["bash"]
```
aarnphm, line 125: For publishing to ghcr I recommend to also setup a actions to build for both arm and amd64

aarnphm (commented):
few tings

```bash

docker run --rm -it -v ${PWD}:/data ghcr.io/ayamir/nvimdots:nightly

docker run --rm -it -v ${PWD}:/data ghcr.io/anymir/nvimdots:stable
```

CharlesChiuGit (approved):
Other than what aarnphm had already pointed out, i think it's go to go as a first draft.

ayamir:
close because it seems meaningless...

PR #1085, opened 2023-11-29
feat(telescope): support live preview of currently selected colorscheme

Cyberczy:
When I use <leader>fc to switch colorschemes,I can't see the change in real time.
So I made some changes

### DIFF
--- a/lua/keymap/tool.lua
+++ b/lua/keymap/tool.lua
@@ -116,7 +116,9 @@ local plug_map = {
 		:with_desc("find: Word in project"),
 	["n|<leader>fe"] = map_cu("Telescope oldfiles"):with_noremap():with_silent():with_desc("find: File by history"),
 	["n|<leader>ff"] = map_cu("Telescope find_files"):with_noremap():with_silent():with_desc("find: File in project"),
-	["n|<leader>fc"] = map_cu("Telescope colorscheme")
+	["n|<leader>fc"] = map_callback(function()
+			require("telescope.builtin").colorscheme({ enable_preview = true })
+		end)
 		:with_noremap()
 		:with_silent()
 		:with_desc("ui: Change colorscheme for current session"),


Cyberczy:
https://github.com/ayamir/nvimdots/assets/109653523/a0c20e9a-3121-4aaf-bffe-adc8331af29f

Cyberczy:
https://github.com/ayamir/nvimdots/assets/109653523/723b8990-07e4-4ae4-ac1e-6b9d79badfec

ayamir (approved):

ayamir (changes requested):
In fact, `alpha.lua` is also needed to change.

Cyberczy:
> 事实上，`alpha.lua`也是需要改变的。

yup, I had that problen too,and I'll try to fix it later

CharlesChiuGit:
nice work!

Cyberczy:
> > In fact, `alpha.lua` is also needed to change.
> 
> I made the same change in `alpha.lua`, but it didn't work


https://github.com/ayamir/nvimdots/assets/109653523/c1188fd6-ca91-408b-b161-84d00f711335

CharlesChiuGit:
I change to lazy event of `telescope.nvim` to `BufWinEnter`, but it wont change the colorscheme in realtime too. I also went dig some docs about `alpha.nvim` and `telescope.nvim`, can't find any useful info too.

image

Cyberczy:
> I change the lazy event of `telescope.nvim` to `BufWinEnter`, but it also wont change the colorscheme in realtime.
> 
> I also went dig some docs about `alpha.nvim` and `telescope.nvim`, can't find any useful info tho.
> 
> image

I had the same problem when using nvim-tree
nvim-tree

Cyberczy:
https://github.com/ayamir/nvimdots/assets/109653523/bdc1f5d9-48ca-45c6-b0ce-f3f712400fb1

In this video, you can clearly see that the README.md file works properly, but the nvim-tree file does not

CharlesChiuGit:
i think it's about `filetype`, for example, alpha's filetype is `alpha` and nvim-tree's filetype is `NvimTree`.

image
image

Cyberczy:
yup, but I'm running out of ideas

CharlesChiuGit:
https://github.com/nvim-telescope/telescope.nvim/blob/84c5a71d825b6687a55aed6f41e98b92fd8e5454/lua/telescope/builtin/__internal.lua#L974-L1019

Jint-lzxy:
> I made the same change in `alpha.lua`, but it didn't work

AFAIU this _is_ the expected behavior under the current implementation. Since these buffers aren't included in the _listed buffer list_, there's no simple way to open such a buffer in another window, and setting the current colorscheme to itself in those buffers (e.g, running `:colorscheme catppuccin` again when the current colorscheme is already `catppuccin`) is usually not supported across implementations. And this is also the behavior when `Telescope` is first opened (because the caret defaults to point to the current active colorscheme.) The following one-liner intuitively demonstrates this:
```sh
nvim unlisted_buffer "+set nobuflisted | Telescope colorscheme enable_preview=true"
```
So I'm kinda on the fence about whether to include this patch in our defaults, because it doesn't work as expected in some (rather common) use cases.

ayamir (approved):

CharlesChiuGit (approved):
I think this is gd enough too.

Jint-lzxy (approved):
Agree with this too lol @Cyberczy thanks for the great work here!

Issue #851, opened 2023-07-10
`null-ls.nvim` has been archived


vollowx:
See [this issue](https://github.com/jose-elias-alvarez/null-ls.nvim/issues/1621)

CharlesChiuGit:
maybe we should roll back to efm? or wait for the fork?

Jint-lzxy:
IMO we can wait for a while to see if there's any progress

Jint-lzxy:
Several interesting points on moving forward & possible perplexity: https://youtu.be/oy_-hQdkoXg.

CharlesChiuGit:
> Several interesting points on moving forward & possible perplexity: https://youtu.be/oy_-hQdkoXg.

I just saw this last night lol.

ayamir:
Don't hurry, waiting a successor for a while

CharlesChiuGit:
https://www.reddit.com/r/neovim/comments/150t159/null_ls_alternatives_solutions

Rinyaresu:
https://github.com/nvimdev/guard.nvim :eyes:

CharlesChiuGit:
> [nvimdev/guard.nvim](https://github.com/nvimdev/guard.nvim) 👀

I'm also looking the status of the plugin closely, since it's by glepnir.

aarnphm:
unpopular opinion: glepnir plugin moves faster than the speed of light, which makes living on the edge nice, with a huge burden to always try to keep up with every plugin changes 😄

vollowx:
which is actually quite popular in my mind for a long time :smile_cat:

ayamir:
https://github.com/stevearc/conform.nvim

vollowx:
The most acceptable way for myself.

aarnphm:
> [stevearc/conform.nvim](https://github.com/stevearc/conform.nvim)

From the quick look of the library, seems like they yet to have support for lazy initialisation. So this probably will need to couple with lsp config 🤔

vollowx:
Also, you may use [nvimtools/none-ls.nvim](https://github.com/nvimtools/none-ls.nvim) as a successor.

CharlesChiuGit:
https://www.reddit.com/r/neovim/comments/1865vvi/formatternvim_or_conformnvim/

Mythos-404:
> https://www.reddit.com/r/neovim/comments/1865vvi/formatternvim_or_conformnvim/

[...]

PR #1088, opened 2023-12-04
feat(plugins): migrate from `null-ls` to `none-ls`

Jint-lzxy:
IMO we can switch to `none-ls` before we decide on the best alternative to `null-ls` b/c it's a drop-in replacement for `null-ls`, and we can continue to receive future bug fixes as well.

### DIFF
--- a/lua/core/settings.lua
+++ b/lua/core/settings.lua
@@ -112,7 +112,7 @@ settings["lsp_deps"] = {
 -- Set the general-purpose servers that will be installed during bootstrap here.
 -- Check the below link for all supported sources.
 -- in `code_actions`, `completion`, `diagnostics`, `formatting`, `hover` folders:
--- https://github.com/jose-elias-alvarez/null-ls.nvim/tree/main/lua/null-ls/builtins
+-- https://github.com/nvimtools/none-ls.nvim/tree/main/lua/null-ls/builtins
 ---@type string[]
 settings["null_ls_deps"] = {
 	"clang_format",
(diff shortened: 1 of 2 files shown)


YuCao16:
Can we consider other alternatives like `conform.nvim`, the advantages of which seem attractive.
https://github.com/ayamir/nvimdots/issues/851#issuecomment-1836305987

ayamir:
I tend to `none-ls` b/c it's maintained by community now.

CharlesChiuGit:
As @Mythos-404 stated:

> After I replaced null-ls with [conform.nvim](https://github.com/ayamir/nvimdots/issues/851) I found that conform.nvim has these advantages over
>
> 1. startup speed improvement
> 2. performance improvement when formatting
> 3. formatting without breaking code folding
> 4. the ability to specify a range of formatting

Could u show us some profile numbers to let we know that's the trade-off?
Or I can do it when I got time.

---

Tho there might be some attractive side of `conform.nvim`, as @YuCao16 said, i dont really see it as a replacement to `null-ls.nvim`. For example, `null-ls` provides "Code actions", "Hover" and so on to be configured; however, i dont see sth relatable in `conform.nvim`? (or did i missed it?)
But i do appreciate the comprehensiveness of `conform.nvim` about formatting.

---

Continuity of the user experiences is also a big part when planing migrations, users with less experiences will pop up with A LOT OF issues. However, if it worth the pain, i think we'll all embrace the migration!

Mythos-404:
> Could u show us some profile numbers to let we know that's the trade-off?

[This is my warehouse.](https://github.com/Mythos-404/nvimdots)

Use `none-ls.nvim` first, after all, `null.nvim` has not been maintained for too long.

Migrating `null-ls.nvim` to `conform.nvim` requires a lot of code changes, which is not cost-effective at the moment, after all, `nvimdots` has a mature `null-ls.nvim` configuration now.

ayamir (approved):

CharlesChiuGit (approved):
we can merge it first.

Issue #981, opened 2023-09-01
Telescope quick fix list cannot work correctly


Saafo:
### Version confirmation

- [X] Confirm

### Following prerequisites

- [X] Confirm

### Not a user config issue

- [X] Confirm

### Neovim version

0.9.0

### Operating system/version

macOS 13.5.1

### Terminal name/version

Warp

### $TERM environment variable

xterm-256color

### Branch info

main (Default/Latest)

### Fetch Preferences

SSH (use_ssh = true)

### How to reproduce the issue

1. Open nvim with Alpha and no other tabs
2. Open Telescope
3. Search something and use `C+q` to add all items in the search result to the quick fix list
4. Double click any item in the quick fix list

### Expected behavior

In step 2, there won't be any error message.
In step 3, the file should display in the main window.

### Actual behavior

In step 2:
<img width="1701" alt="image" src="https://github.com/ayamir/nvimdots/assets/46736350/cc0b177d-f1ba-4121-b5f1-6abbeb2f2758">

In step 3:
<img width="1684" alt="image" src="https://github.com/ayamir/nvimdots/assets/46736350/3e76adb2-38bd-46e4-8cf9-ca0a2cbbcf86">

ayamir:
The error of `indent-blankline` is fixed in `0ce5f35ef07bdff93bd31e8a979dc9801d0e0f2d`.

The second issue maybe have to do with `alpha-nvim` itself. I found a similar thread in its discussions: https://github.com/goolord/alpha-nvim/discussions/195

ayamir:
I fixed it by using [nvim-bqf](https://github.com/kevinhwang91/nvim-bqf) and it's a awesome plugin.

Saafo:
Thanks I'll try it!!

Saafo:
I've tried `nvim-bqf` and it's pretty good! Thanks for the recommendation! And I don't care about the original secondary issue since `nvim-bqf` works greatly. If nobody else cares, feel free to close this issue.

PR #1091, opened 2023-12-12
feat(plugins): add `nvim-bqf`

Jint-lzxy:
IMO we could think about bringing `nvim-bqf` back. Not only does it make nvim's quickfix mode a breeze to use, but it also sidesteps some of the issues we had before, like #981. Plus, it's a plugin that loads on demand, so it shouldn't cause any side effects 😄


ayamir (approved):
Agree.

CharlesChiuGit (approved):
LGTM

PR #1098, opened 2023-12-14
feat: add plugin zeioth/garbage-day.nvim.

jczhang02:
Not sure the impact of the plugin. Maybe have a discussion?

### DIFF
--- a/lua/modules/plugins/tool.lua
+++ b/lua/modules/plugins/tool.lua
@@ -63,6 +63,10 @@ tool["gelguy/wilder.nvim"] = {
 	config = require("tool.wilder"),
 	dependencies = { "romgrk/fzy-lua-native" },
 }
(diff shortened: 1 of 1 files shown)


CharlesChiuGit:
What's the difference between https://github.com/hinell/lsp-timeout.nvim and https://github.com/Zeioth/garbage-day.nvim?
I cant really tell the advantages from both of it's doc.

jczhang02:
you are right! My mistake! I will close this pr.

CharlesChiuGit:
What? XDDD
We haven't adopt one of those yet, but i think it's nice to have sth similar.
i just want to know which one is better lol

jczhang02:
I mean that i will use this plugin and test the two plugins lol.

CharlesChiuGit:
ah, ok lol.
than i'll close the pr for now.

PR #1114, opened 2023-12-25
feat(script): skip dependency installation by default

Jint-lzxy:
IMO the default choice for dependency installation in `install.ps1` is a bit misleading (see Discussion #1106). afaik most users won't even "visit" the dynamic provider feature of Neovim, and those who know what they're doing can still continue with the dependency installation by typing `Y/y`. So I think a more reasonable default should be `N` (i.e. skipping provider installation altogether.)


CharlesChiuGit (approved):
lgtm

ayamir (approved):
Agree.

Issue #1135, opened 2024-01-07
Use mouse shut down accidentally.


Cheny-chui:
### Version confirmation

- [X] Confirm

### Following prerequisites

- [X] Confirm

### Not a user config issue

- [X] Confirm

### Neovim version

NVIM v0.9.4

### Operating system/version

macOS 13.5.2 (22G91)

### Terminal name/version

iTerm2 Build 3.4.23

### Branch info

main (Default/Latest)

### Fetch Preferences

HTTPS (use_ssh = false)

### How to reproduce the issue

1. open a project.
2. open at least two buffer, e.g. file1, file2
3. open the file explorer 
4. use **mouse** to close the **rightmost** buffer (click on the x symbol)

### Expected behavior

expect only one buffer close

### Actual behavior

the whole nvim shut down

Cyberczy:
look this https://github.com/ayamir/nvimdots/pull/1136

PR #1136, opened 2024-01-07
fix(bufferline): use `BufDel %d` instead of `bdelete! %d`

Cyberczy:
When I use the close button on bufferline, it often exits abnormally.
After using BufDel %d instead of bdelete! %d, this situation no longer occurs.


Cyberczy:
fix https://github.com/ayamir/nvimdots/issues/1135

CharlesChiuGit (approved):
can confrim. it works.

ayamir (approved):

Issue #1132, opened 2024-01-06
[deps] rust-tools.nvim has been archived.


ClSlaid:
### Feature description

Luckily, there is still a heavily customized fork of it.
https://github.com/mrcjkb/rustaceanvim
I wonder if we should move from `rust-tools` to it.

CharlesChiuGit:
since most of us don't write rust, maybe a pr will be nice?

PR #1140, opened 2024-01-08
patch: remove rust-tools.nvim

ClSlaid:
Rust-tools.nvim has been archived.

Use rustaceanvim instead. Note that this plugin requires a version of nvim >= 0.9.

related: #1132


ayamir:
Two rust_analyzer were spawned for me:
image

ClSlaid:
> Two rust_analyzer were spawned for me:

It could happen if you have both Mason managed and rustup managed rust-analyzer installed.

Please remove one of them, and then restart the lsp.

PR #1143, opened 2024-01-10
fix(bufferline): avoid unnecessary user interaction

Jint-lzxy:
This is a follow-up fix for #1136. It prevents `bufferline` from tripping over errors thrown by `nvim-bufdel`, and is also keeping up with the upstream close command (`bdelete!`)


CharlesChiuGit (approved):
LGTM

PR #1145, opened 2024-01-10
fix(rust): respect user configs

Jint-lzxy:
This PR does three things:

- Allows users to overwrite the init options of `rustaceanvim` thru their custom configs;

- It gives a heads-up to users that if they're still manually setting up `rust_analyzer`, [this will result in two identical servers being spawned](https://github.com/ayamir/nvimdots/pull/1140#issuecomment-1881190708);

- DAP-related settings are disabled by default to avoid potential conflicts with previous user configurations. IMO since we already have [configs for `codelldb`](https://github.com/ayamir/nvimdots/blob/f2cf03b254c21a69d3654c2c05014cd411ac14e2/lua/modules/configs/tool/dap/clients/codelldb.lua#L46) that can be used, users can simply tweak those configs on their own. But I'm not 2 sure about this, as I personally don't do much rust development. What do u guys think @ClSlaid @CharlesChiuGit @ayamir?

### DIFF
--- a/lua/modules/configs/completion/mason-lspconfig.lua
+++ b/lua/modules/configs/completion/mason-lspconfig.lua
@@ -28,6 +28,23 @@ M.setup = function()
 	---A handler to setup all servers defined under `completion/servers/*.lua`
 	---@param lsp_name string
 	local function mason_lsp_handler(lsp_name)
+		-- rust_analyzer is configured using mrcjkb/rustaceanvim
+		-- warn users if they have set it up manually
+		if lsp_name == "rust_analyzer" then
+			local config_exist = pcall(require, "completion.servers." .. lsp_name)
+			if config_exist then
+				vim.notify(
+					[[
+`rust_analyzer` is configured independently via `mrcjkb/rustaceanvim`. To get rid of this warning,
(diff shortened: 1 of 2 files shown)


ClSlaid (approved):

ayamir (commented):
LGTM

ayamir (approved):

CharlesChiuGit (approved):
seems reasonable

yyb196:
I use the current master branch, and I also have two rust-analyzer LSP server, I don't get the WARN notification this pr add, please help, how can I solve this

image

yyb196:
after I remove rust_analyzer.lua in lsp-nvimconfig plugin, it works, only one LSP server left for rust buffer, I have no idea:
image
image

yyb196:
@CharlesChiuGit thanks for your feed back, I have backport the pr you mentioned a couple dayes ago, still don't work. 

or is there a command or operation needed to make new config work after changing config?

CharlesChiuGit:
have u do `:Lazy sync` to update all the plugins?

CharlesChiuGit:
@yyb196 have u do `:Lazy sync` to update all the plugins?

PR #1162, opened 2024-02-03
fix(scrollview): omit the `scrollview_` prefix

Cyberczy:
https://github.com/dstein64/nvim-scrollview?tab=readme-ov-file#lua-example
A Lua `setup()` function is provided for convenience, to set globally scoped options (the 'scrollview_' prefix is omitted).


CharlesChiuGit (approved):
LGTM

ayamir (approved):

PR #1165, opened 2024-02-05
feat: make `lazy` option configurable.

ayamir:
Previous implementation doesn't consider configurable lazy options like `lazy = false`, which can't make some plugins function works normally such as `autoload = true` for `persisted.nvim`. I only make this option configurable because the modification of other options like `event`, `cmd`, `cond`... etc, should be done by ourselves rather than users.


ayamir:
My fault, this feature can be impl by add config directly in `user/plugins/<scope>.lua` in fact lol.

PR #1173, opened 2024-02-11
chore(ci): bump deps versions

Cyberczy:


Cyberczy:
Happy lunar new year 🎉

ayamir (approved):

Cyberczy:
Perhaps we should use dependabot to update deps versions.
This way we can avoid forgetting to update these deps versions.
see [https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates](https://docs.github.com/en/code-security/dependabot/dependabot-security-updates/configuring-dependabot-security-updates)
How do you think? @ayamir

CharlesChiuGit:
i dont think using a dependabot is a good idea.
we might have [supply chain attack](https://www.cloudflare.com/learning/security/what-is-a-supply-chain-attack/) issue.
[it's very own security news.](https://www.darkreading.com/application-security/supply-chain-attackers-escalate-with-github-dependabot-impersonation)

PR #1174, opened 2024-02-12
feat(ci): use dependabot to bump deps versions

Cyberczy:


ayamir (approved):

CharlesChiuGit:
i dont think using a dependabot is a good idea.
we might have [supply chain attack](https://www.cloudflare.com/learning/security/what-is-a-supply-chain-attack/) issue.
[it's very own security news.](https://www.darkreading.com/application-security/supply-chain-attackers-escalate-with-github-dependabot-impersonation)

ayamir:
Aha, it's indeed a potential risk. Maybe we can revert this commit.

Cyberczy:
agree

PR #1197, opened 2024-03-13
feat(ui): enable global statusline, disable all extension.

ayamir:
Global statusline is more friendly for multiple window splits.
Besides, I disabled all extension which looks not suitable for global statusline.

image


CharlesChiuGit (approved):
LGTM!

PR #1206, opened 2024-03-22
delete duplicate config about resize windows

HATTER-LONG:
RT


ayamir:
The reason why I left them in `mappings.lua` is for the situation that `smart-splits.nvim` removed should also has the mappings to resize window.

Issue #1211, opened 2024-03-31
Improve NixOS support


misumisumi:
### Feature description

I'm glad to have contributed to the support for NixOS last time.
Since then, we have been able to operate it stably for several months.
However, previous support did not provide an option to place `lazy-lock.json` under `.config/nvim`.
This is because NixOS places it as read-only, which means it cannot be updated with `:Lazy`.
Fortunately, we were able to create a workflow that updates `lazy-lock.json` by following upstream in a forked repository based on `update_lockfile.yml`, so the above problem will be alleviated.
I provide bind support in terms of the NixOS ethos of strongly anchoring the environment.

Also, last time it was provided as `nixosModules`, but since it actually only works with home-manager, we changed it to `homeManagerModules`. (breaking change)
This is because I didn't know about the existence of `homeManagerModules` last time.

### Additional information

- example merge upstream and update lazy-lock workflow
In the example, I create a PR instead of just pushing it to `main`

```yaml
name: auto merge upstream
on:
  workflow_dispatch: # allows manual triggering
  schedule:
    - cron: "0 19 * * 5" # runs weekly on Saturday at 04:00 (JST)

permissions:
  contents: write
  pull-requests: write
jobs:
  auto-merge-upstream:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
        with:
          fetch-depth: 0

      - uses: andstor/file-existence-action@v3
        id: check_lockfile
        with:
          files: "lazy-lock.json"

      - name: Merge upstream
        id: merge-upstream
        continue-on-error: true
        run: |
          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
          git config --global user.name "github-actions[bot]"

          git remote add upstream https://github.com/ayamir/nvimdots.git
          git fetch upstream
          git merge upstream/main --allow-unrelated-histories --no-commit --no-ff

          git reset HEAD lazy-lock.json
          git reset HEAD .github/workflows/update_lockfile.yml

          git checkout lazy-lock.json
          git checkout .github/workflows/update_lockfile.yml

      - name: Run count-new-commits
        run: |
          echo "NEW_COMMIT_COUNT=$(git log --oneline --since '24 hours ago' --perl-regexp --author='^((?!github-actions).*)$' | wc -l)" >> "$GITHUB_ENV"

      - uses: rhysd/action-setup-vim@v1
[...]

PR #1212, opened 2024-03-31
Improve nixos

misumisumi:
#1211 

- Support for placing `lazy-config.json` under `.config/nvim`
- Change from `nixosModules` to `homeManagerModules` (BREAKING CHANGE)


ayamir:
Thanks for your continuous contributions! I don't have `nix` or `nixos` environment so can't test it. Could you please check it? @aarnphm @CharlesChiuGit

nixos/neovim/default.nix
```diff
@@ -89,32 +94,21 @@ in
...
-        name = "nvim-depends-include";
-        paths = splitString ":" (makeIncludePath build-dependent-pkgs);
-        extraPrefix = "/lib/nvim-depends/include";
-        ignoreCollisions = true;
-      };
-      nvim-depends-pkgconfig = pkgs.buildEnv {
-        name = "nvim-depends-pkgconfig";
-        paths = splitString ":" (makePkgConfigPath build-dependent-pkgs);
-        extraPrefix = "/lib/nvim-depends/pkgconfig";
+      neovim-build-deps = pkgs.buildEnv {
```
aarnphm, line 97: qq: why do we have to build neovim from scratch again?

Can we utilize something like neovim-nightly or sth?

CharlesChiuGit: i can add that, but i'm half way on completing my nix config.
i'll check this pr once i'm done.

aarnphm: sg. I have yet to migrate all my neovim to nix yet so haven't tested it extensively.

ayamir:
Does this PR work as expect?

nixos/neovim/default.nix
CharlesChiuGit: i guess i'm wrong, it's not the same as adding overlays in home manager's `home.nix`.

CharlesChiuGit:
@misumisumi 
why do we need to include `dotnet`? is that necessary?

misumisumi:
This is a little difficult when using dotnet in a NIX environment, so we made it modular to make it easier to configure.
Actually, there is a procedure in A, so there is no problem even if it is not included here.

PR #1219, opened 2024-04-05
feat(cmp): more completion candidates from opened buffers.

mjkx5:
using that, buffer completion include all buffer, not only one buffer.

https://github.com/hrsh7th/cmp-buffer

get_bufnrs (type: fun(): number[])
Default: function() return { vim.api.nvim_get_current_buf() } end

A function that specifies the buffer numbers to complete.

### DIFF
--- a/lua/modules/configs/completion/cmp.lua
+++ b/lua/modules/configs/completion/cmp.lua
@@ -172,7 +172,7 @@ return function()
 			{ name = "orgmode" },
 			{
 				name = "buffer",
-				opts = {
+				option = {
 					get_bufnrs = function()
 						return vim.api.nvim_list_bufs()
 					end,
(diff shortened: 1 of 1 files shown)


ayamir:
What's the effect?

mjkx5:
Screenshot_06-Apr_16-55-40_kitty
Screenshot_06-Apr_16-59-30_kitty
It is the first picture which is add this config.
It is the second picture which is not add the config.

Default: vim.api.nvim_get_current_buf() -> only read current buffer to completion

the modified config -> reading all buffer to completion.

ayamir:
Understood. Useful feature!

Jint-lzxy:
Hmm, but wouldn't this potentially lead to some inappropriate suggestions, especially when buffers are open for different file formats? \[e.g., `cmp-buffer` might suggest `@autoreleasepool` in a buffer containing C source code (which is invalid) because the project also includes Objective-C files.\] Plus, there could be performance implications when too many buffers are open 🤔

ayamir:
IMO the range of candidates will decrease during the input process. This function can be seen as a supliment of lsp suggestion, specically for the text input situation.

PR #1226, opened 2024-04-09
feat!: improve go development experience

### LINKED ISSUE #1207: why use vim-go instead of go.nvim
### Feature description

- `go.nvim` is **github.com/ray-x/go.nvim**
- `vim-go` is **github.com/fatih/vim-go**

go.nvim offers more modern features and configurations than vim-go, and has an active community. Therefore, it is recommended to update the development plug-in for go to "ray-x/ go.nvim ".

### Additional information

_No response_

ayamir:
This PR aims to improve go development workflow, it will close #1207.
Welcome for gophers like @kincoy @mjkx5 to review this PR, join discussion and make contribution!


lua/keymap/completion.lua
```diff
@@ -50,6 +51,10 @@ function mapping.lsp(buf)
 		["n|gd"] = map_cr("Glance definitions"):with_silent():with_buffer(buf):with_desc("lsp: Preview definition"),
 		["n|gD"] = map_cr("Lspsaga goto_definition"):with_silent():with_buffer(buf):with_desc("lsp: Goto definition"),
 		["n|gh"] = map_cr("Glance references"):with_silent():with_buffer(buf):with_desc("lsp: Show reference"),
+		["n|gi"] = map_cu("lua vim.lsp.buf.implementation()")
```
Jint-lzxy, line 54:
```suggestion
		["n|gm"] = map_cr("Glance implementations")
```
(Glace provides this [feature](https://github.com/DNLHC/glance.nvim?tab=readme-ov-file#commands))

qq: Is using `gi` really a good choice here? Because at least in my workflow, I sometimes need it:
```console
							*gi*
gi			Insert text in the same position as where Insert mode
			was stopped last time in the current buffer.
			This uses the |'^| mark.  It's different from "`^i"
			when the mark is past the end of the line.
			The position is corrected for inserted/deleted lines,
			but NOT for inserted/deleted characters.
			When the |:keepjumps| command modifier is used the |'^|
			mark won't be changed.
```

ayamir:
IMO we can add Go related buffer keymaps like `GoIfErr`, `GoModTidy`, `GoTest` etc...

ayamir:
I think we'd better avoid keymap intruduction and should leave this to user. So it ready to review now.

lua/keymap/completion.lua
```diff
@@ -4,7 +4,7 @@ local map_cmd = bind.map_cmd
 local map_callback = bind.map_callback
 
 local plug_map = {
-	["n|<A-f>"] = map_cmd("<Cmd>FormatToggle<CR>"):with_noremap():with_desc("Formater: Toggle format on save"),
+	["n|<A-f>"] = map_cmd("<Cmd>FormatToggle<CR>"):with_noremap():with_desc("Formatter: Toggle format on save"),
```
Jint-lzxy, line 7:
```suggestion
	["n|<A-f>"] = map_cmd("<Cmd>FormatToggle<CR>"):with_noremap():with_desc("formatter: Toggle format on save"),
```

Changed to lowercase for better consistency :D

lua/keymap/completion.lua
```diff
@@ -50,6 +50,10 @@ function mapping.lsp(buf)
 		["n|gd"] = map_cr("Glance definitions"):with_silent():with_buffer(buf):with_desc("lsp: Preview definition"),
 		["n|gD"] = map_cr("Lspsaga goto_definition"):with_silent():with_buffer(buf):with_desc("lsp: Goto definition"),
 		["n|gh"] = map_cr("Glance references"):with_silent():with_buffer(buf):with_desc("lsp: Show reference"),
+		["n|gm"] = map_cr("Glance implementations")
+			:with_silent()
+			:with_buffer(buf)
+			:with_desc("lsp: Goto implementation"),
```
Jint-lzxy, line 56:
```suggestion
			:with_desc("lsp: Show implementation"),
```

I think Glance doesn't directly lead users to the implementation.

lua/modules/configs/lang/go-nvim.lua
```diff
@@ -0,0 +1,24 @@
...
+		lsp_document_formatting = true,
+		lsp_codelens = false,
+		-- disable inlay hints by default
+		lsp_inlay_hints = { enable = false },
+		-- show failed test info in trouble window
+		trouble = true,
+		-- show passed test info in notify message
+		run_in_floaterm = false,
+		-- gopls config
+		lsp_cfg = require("modules.configs.completion.servers.gopls"),
```
Jint-lzxy, line 22: I'm still going through everything, but is the setup for `gopls` supposed to be like this? I think if the plugin sets it up too, we might end up with duplicate `gopls` instances - mason-lspconfig would also set up the server using `servers.gopls.lua` provided the plugin doesn't do any runtime checks.  Should we give users the same warning [as we did for rustaceanvim](https://github.com/ayamir/nvimdots/blob/651a25962e2db53933cae8c0aeb917916e1a4174/lua/modules/configs/completion/mason-lspconfig.lua#L31-L46)?

lua/modules/plugins/lang.lua
```diff
@@ -8,11 +8,16 @@ lang["kevinhwang91/nvim-bqf"] = {
...
-lang["fatih/vim-go"] = {
+lang["ray-x/go.nvim"] = {
 	lazy = true,
-	ft = "go",
-	build = ":GoInstallBinaries",
-	init = require("lang.vim-go"),
+	dependencies = { -- optional packages
+		"ray-x/guihua.lua",
+		"neovim/nvim-lspconfig",
+		"nvim-treesitter/nvim-treesitter",
```
Jint-lzxy, line 16: I think we can remove these two since the config should suffice without them.

lua/modules/plugins/lang.lua
```diff
@@ -8,11 +8,16 @@ lang["kevinhwang91/nvim-bqf"] = {
...
+	},
+	ft = { "go", "gomod" },
+	build = ':lua require("go.install").update_all_sync()', -- if you need to install/update all binaries
```
Jint-lzxy, line 19:
```suggestion
	build = ":GoInstallBinaries",
```

I think [`:GoInstallBinaries`](https://github.com/ray-x/go.nvim?tab=readme-ov-file#go-binaries-install-and-update) should be sufficient in this case since all the other packages are managed by Mason. What's the purpose of calling this function here?

Jint-lzxy:
@ayamir The main motivation behind the last three commits is that I see `go.nvim` more as an enhancement layer for Go development, rather than a complete "dev bundle" akin to those in VSCode. So I've reintroduced some basic features like LSP and formatting support to our "native" way of handling them, leaving the "fancy stuff" to `go.nvim` (I've also integrated some good stuff from `go.nvim`'s default LSP config tho). IMO this approach avoids introducing breaking changes while also giving users the option to completely remove this plugin without worrying about a dysfunctional Go development environment. But it's just a suggestion! I'm open to reverting these changes if needed lol

ayamir:
LGTM 👍

Jint-lzxy (approved):

PR #1229, opened 2024-04-11
fix(options): remove options managed by `vim-sleuth`

Jint-lzxy:
Since all tab-related stuff is now managed by `vim-sleuth`, I think it is better to remove the relevant options to avoid confusion.


ayamir (approved):
Make sense

PR #1233, opened 2024-04-18
fix(lualine): disable default separators for custom separator

Jint-lzxy:
I know the commit message might be a bit difficult to grasp, but I'm struggling to find a clearer way to describe it :( Essentially, this commit ensures that when our custom separator is used, default separators (as defined by `section_separators` and `component_separators`) are not added to the statusline. This prevents certain extreme scenarios where the default separator cannot be concealed, such as when in terminal mode with `:ToggleTerm direction=horizontal`.


ayamir (approved):
LGTM

PR #1236, opened 2024-04-19
fix: inlay_hint error

Mythos-404:
After the commit [`f1dfe32`](https://github.com/neovim/neovim/commit/f1dfe32bf5552197e0068298b0527526a4f918b1) in `neovim`, the function `vim.lsp.inlay_hint.enable(buffer_number, enable)` was changed to `vim.lsp.inlay_hint.enable(enable: boolean, filter: table)`, where the original approach would lead to an error.

**I suspect there could be further changes to this API.**


ayamir (approved):
LGTM

CharlesChiuGit:
cc https://neovim.io/doc/user/lsp.html#lsp-inlay_hint

PR #1237, opened 2024-04-20
fix: prevent horizontal terminal from obscuring `nvim-tree`

ayamir:
IMO this is more reasonable behavior when nvim-tree opened as some users discussed in https://github.com/akinsho/toggleterm.nvim/issues/97

### DIFF
--- a/lua/modules/configs/tool/toggleterm.lua
+++ b/lua/modules/configs/tool/toggleterm.lua
@@ -3,16 +3,25 @@ return function()
 		-- size can be a number or function which is passed the current terminal
 		size = function(term)
 			if term.direction == "horizontal" then
-				return 15
+				return vim.o.lines * 0.40
 			elseif term.direction == "vertical" then
 				return vim.o.columns * 0.40
 			end
 		end,
-		on_open = function()
+		on_open = function(term)
 			-- Prevent infinite calls from freezing neovim.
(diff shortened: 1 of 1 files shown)


Jint-lzxy (approved):
LGTM! In case anyone's wondering why this isn't the default, [akinsho's comment](https://github.com/akinsho/toggleterm.nvim/pull/388#issuecomment-1425935832) might shed some light on that:

> The main reason I chose the current strategy for toggleterm is because if you split per window then adding more splits to an already open toggleterm doesn't work as well because there isn't much space for new splits if opened in window that happens to be small. Also I don't want the terminals to seem to be scoped to the window. That isn't the point they are general terminals not per window terminals.
>
> Also it isn't really reasonable to expect neovim to work like other IDEs because it isn't designed like other IDEs for example despite it being kind of annoying special filetree windows are just windows there is no reason/ no way for me to easily open the terminal but not also open it under the file tree because neovim doesn't special case these buffers.
>
> Anyway as it stands this PR is more likely to break things and the only proper way to really "respect" sidebars would be to have either builtin respect for sidebars when splitting which I don't think the nvim team are likely to add anytime soon or adding complicated logic to check if there is some sort of specified file tree in a window by looping through all the windows then if so you split like it as it does before the switch to the file tree window and use something like `wincmd H` to re-extend it's size

PR #1238, opened 2024-04-20
pref: disable `<C-e>` for open-in-place inside nvim-tree

ayamir:
This PR disable `<C-e>` in `nvim-tree` buffer which will open file in place originally. IMO just leaving it to scroll is better and we usually don't want to open a file in such a narrow buffer.

### DIFF
--- a/lua/modules/configs/tool/nvim-tree.lua
+++ b/lua/modules/configs/tool/nvim-tree.lua
@@ -18,8 +18,7 @@ return function()
 		sort_by = "name",
 		sync_root_with_cwd = true,
 		on_attach = function(bufnr)
-			local api = require("nvim-tree.api")
-			api.config.mappings.default_on_attach(bufnr)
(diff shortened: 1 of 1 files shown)


lua/modules/configs/tool/nvim-tree.lua
```diff
@@ -17,6 +17,11 @@ return function()
 		respect_buf_cwd = false,
 		sort_by = "name",
 		sync_root_with_cwd = true,
+		on_attach = function(bufnr)
+			local api = require("nvim-tree.api")
+			api.config.mappings.default_on_attach(bufnr)
```
Jint-lzxy, line 22:
```suggestion
			require("nvim-tree.api").config.mappings.default_on_attach(bufnr)
```

I think we can just merge these two lines without needing another local variable. Other than that this PR sounds reasonable to me.

Jint-lzxy (approved):

PR #1240, opened 2024-04-22
feat(catppuccin): remove all styles of `@property`

Jint-lzxy:
Based on the [upstream spec](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#types), `@property` is typically associated with the "key" in key/value pairs. So IMO since we don't give special treatment to the "values" (such as assigning them an "italic" property), we should treat the "keys" in the same manner as well.


CharlesChiuGit (approved):
lgtm

ayamir (approved):

Issue #1203, opened 2024-03-19
dashboard_image error


jwangtobeno1:
### Version confirmation

- [X] Confirm

### Following prerequisites

- [X] Confirm

### Not a user config issue

- [X] Confirm

### Neovim version

NVIM v0.9.5

### Operating system/version

Arch WSL2

### Terminal name/version

WSL2

### Branch info

main (Default/Latest)

### Fetch Preferences

SSH (use_ssh = true)

### How to reproduce the issue

The following problem occurs when I use this  [method](https://github.com/ayamir/nvimdots/wiki/Issues#change-dashboard-startup-image)  to change dashboard image 

![image](https://github.com/ayamir/nvimdots/assets/69038155/0afe1fe9-57b8-48ef-a5e5-68e1edc8e953)

The Follow is my settings:

![image](https://github.com/ayamir/nvimdots/assets/69038155/0a1c3d93-7a82-4687-9946-72f8604db8da)

### Expected behavior

only show my image

ayamir:
```lua
settings["dashboard_image"] = function()
	return {
		-- your ascii image
	}
end
```

jwangtobeno1:
ok，it works

> ```lua
> settings["dashboard_image"] = function()
> 	return {
> 		-- your ascii image
> 	}
> end
> ```

PR #1244, opened 2024-04-25
fix(dashboard_image): replace predefined dashboard_image with user defined image instead of extending it

csyJoy:


CharlesChiuGit (approved):
lgtm

ayamir (approved):

ayamir:
LGTM and there is another solution: https://github.com/ayamir/nvimdots/issues/1203

Issue #1246, opened 2024-04-28
golang code hints default selected location is incorrect


JohnWickzzz:
### Version confirmation

- [X] Confirm

### Following prerequisites

- [X] Confirm

### Not a user config issue

- [X] Confirm

### Neovim version

NVIM v0.9.5

### Operating system/version

macOS 12.7.3

### Terminal name/version

iTerm2 Build 3.5.0beta24

### Branch info

main (Default/Latest)

### Fetch Preferences

HTTPS (use_ssh = false)

### How to reproduce the issue

![image](https://github.com/ayamir/nvimdots/assets/12469651/b9eae9a3-1416-43c6-a22a-44ad2ab3dae8)

1. Use neovim open a golang file.
2. Input 'go' char, wait the code hints.
3. Default selection is not the first one.

### Expected behavior

Code hints should select the first one, or like other language not select anyone.
![image](https://github.com/ayamir/nvimdots/assets/12469651/c771e820-1871-4b73-9db9-40c799a59a47)

PR #1247, opened 2024-04-28
fix: set cmp preselect to none.

ayamir:
#1246

### DIFF
--- a/lua/modules/configs/completion/cmp.lua
+++ b/lua/modules/configs/completion/cmp.lua
@@ -66,7 +66,7 @@ return function()
 
 	local cmp = require("cmp")
 	require("modules.utils").load_plugin("cmp", {
-		preselect = cmp.PreselectMode.Item,
+		preselect = cmp.PreselectMode.None,
 		window = {
 			completion = {
 				border = border("PmenuBorder"),


JohnWickzzz:
image

After update preselect like this and add completeopt param. Simple test of java/go/lua file, now it can select the first item. But I don't know if it can run normally in all places.

lua/modules/configs/completion/cmp.lua
```diff
@@ -66,7 +66,7 @@ return function()
 
 	local cmp = require("cmp")
 	require("modules.utils").load_plugin("cmp", {
-		preselect = cmp.PreselectMode.Item,
+		preselect = cmp.PreselectMode.None,
```
Jint-lzxy, line 69: qq: I think given the current changes, we could introduce a new keymap to commit the default selection, similar to the following using `<C-CR>` to replicate the original behavior:
```lua
["<C-CR>"] = cmp.mapping.confirm({ select = true, behavior = cmp.ConfirmBehavior.Insert }),
```

ayamir: This PR mainly fixed the cmp preselect item inconsistency when editing different lang codes. Now preselect is set to NONE. IMO it has nothing to do with a new keymap to commit the default preselect. Additionly, seems my terminal can't distinguish `C-CR` from `CR`.

Jint-lzxy: Agreed.

PR #1250, opened 2024-04-29
Fix/nixos

misumisumi:
I found two problems with the previous PR.
One is with errors in the current release version of NixOS (23.11), another is with `luarocks` not being set in the path.
I will therefore fix these issues.
For verification purposes, we have also created a minimal `devshell` environment that runs the settings of this repository.
This will make it possible to verify using a `nix` environment, even if NixOS or HomeManager are not in place.
Since `NVIM_APPNAME` is used, there is no impact on the existing environment.
If there is a slight problem, you may need to put in the effort to keep `flake.lock` up to date.
This can be automated with github actions, just like updating `lazy-lock.json`.
Also, since it doesn't relate to the module part, you may want to update it only when necessary.
I would like to add github actions if necessary.

@CharlesChiuGit 
If possible, could you please confirm if this `devshell` works in an `arm Mac` environment?
It will start with `nix develop`.


nixos/neovim/default.nix
```diff
@@ -114,92 +112,92 @@ in
...
+            tree-sitter
+          ]
+          ++ optionals cfg.withBuildTools [
+            cargo
+            clang
+            cmake
+            gcc
+            gnumake
+            go
             luarocks
```
CharlesChiuGit, line 153: it's better to use `lua51Packages.luarocks` to align with the version of nvim-embeded lua

nixos/neovim/default.nix
```diff
@@ -114,92 +112,92 @@ in
...
-                name = "stack";
-                text = ''
-                  exec "${pkgs.stack}/bin/stack" "--extra-include-dirs=${config.home.profileDirectory}/lib/nvim-depends/include" "--extra-lib-dirs=${config.home.profileDirectory}/lib/nvim-depends/lib" "$@"
-                '';
-              })
-              (haskellPackages.ghcWithPackages (ps: cfg.extraHaskellPackages pkgs.haskellPackages))
-            ];
+        withNodeJs = true;
+        withPython3 = true;
+        withRuby = true;
```
CharlesChiuGit, line 138: i think it's better to set `withRuby` to false, since nvim's ruby plugin is not used in our config.

CharlesChiuGit (changes requested):

CharlesChiuGit:
> If possible, could you please confirm if this devshell works in an arm Mac environment?

ys, it does.
image

CharlesChiuGit:
> This can be automated with github actions, just like updating `lazy-lock.json`.
Also, since it doesn't relate to the module part, you may want to update it only when necessary.
I would like to add github actions if necessary.

agree, it would be nice.

misumisumi:
I tried excluding unsupported packages, will this work?

misumisumi:
> This can be automated with github actions, just like updating lazy-lock.json.
Also, since it doesn't relate to the module part, you may want to update it only when necessary.
I would like to add github actions if necessary.

I had raised an ambiguous problem.,
Is this a positive statement, whether it is better to add a workflow or not?

CharlesChiuGit:
> Is this a positive statement, whether it is better to add a workflow or not?

ys, add a github action would be nice.

nixos/neovim/default.nix
```diff
@@ -69,30 +67,29 @@ in
...
-        ]
-        ++ cfg.extraDependentPackages;
+      build-dependent-pkgs = with pkgs; builtins.filter (package: !package.meta.unsupported) [
+        # manylinux
+        acl
+        attr
+        bzip2
+        curl
+        glibc
+        libgcc.lib
```
CharlesChiuGit, line 77: when i tried to do `nix develop --show-trace`, at the very end, it shows:
```
       … while selecting an attribute
         at /nix/store/mmcw46lk4pfgjz2pys7rhc6ra4mqxk88-source/nixos/neovim/default.nix:77:9:
           76|         glibc
           77|         libgcc.lib
             |         ^
           78|         libsodium

       error: expected a set but found null: null
```
and i couldn't find `libgcc.lib` and `stdenv.cc.cc` in nixpkg, but only `libgcc` and `stdenv`. Is this some nix syntax i don't know or we actually dont need the `.lib` and `.cc.cc` part.

CharlesChiuGit (changes requested):

nixos/neovim/default.nix
misumisumi: `.lib` and `.cc.cc` are necessary.
These are like `*-dev` on ubuntu.
I checked and found that libgcc does not exist on darwin.
This is due to the use of `clang` on darwin.
Therefore, I made it so that it references `stdenv.cc.cc.lib` instead of `libgcc.lib`.
`stdenv.cc.cc.lib` is a wrapper for `libgcc.lib` on linux and `clang.lib` on darwin.

CharlesChiuGit: tks for explanation!

CharlesChiuGit: it works now, nice!
image

flake.nix
```diff
@@ -4,11 +4,76 @@
...
+                And seted NVIM_APPNAME=nvimdots, so neovim will put file under "''${XDG_xxx_HOME}"/nvimdots
+
+                $(type -p menu &>/dev/null && menu)
+              '';
+              startup = {
+                mkNvimDir = {
+                  text = ''
+                    mkdir -p "''${XDG_CONFIG_HOME}"/nvimdots
+                    for path in init.lua lua snips tutor; do
+                      ln -rsf "''${path}" "''${XDG_CONFIG_HOME}"/nvimdots/
```
CharlesChiuGit, line 56: `-r` flag doesn't exists on darwin too.
image

image
is this flag necessary?

misumisumi: Ah.
It was still the old setting.
I'll fix it.

misumisumi: As mentioned earlier, nixpkgs' stdenv uses gcc on Linux and clang on Darwin, but what about `mason.nvim`?
Do I need to change any environment variables?

misumisumi: Will this work?

CharlesChiuGit: seems fine on my end.
image

misumisumi: Did you also build the package?

CharlesChiuGit: ys, all LSP managed by mason works fine.
image

image

misumisumi:
I try not to place lazy-lock.json under $XDG_CONFIG_HOME/nvimdots, but should I place it there?
I have been avoiding the lock file because it seems to be updated on the first startup.

CharlesChiuGit:
`copilot.lua` and `nvim-nio` keeps failing to checkout no matter what i do.
image
i tried remove those two plugins in `~/.local/share/nvimdots/site/lazy` and let `lazy` reinstall it, but it just keeps failing. I cant tell what could cause the issue.

CharlesChiuGit:
> I have been avoiding the lock file because it seems to be updated on the first startup.

interesing, this issue doesn't seem to happen on my end.

> I try not to place lazy-lock.json under $XDG_CONFIG_HOME/nvimdots, but should I place it there?

i'm fine placing the lock file in or out, as long as it got managed by the user or  via our nvim config. Personally i like to place the lock file in `$XDG_CONFIG_HOME/nvimdots`, so it's faster for me to debug if needed.

Jint-lzxy:
> i tried remove those two plugins in `~/.local/share/nvimdots/site/lazy` and let `lazy` reinstall it, but it just keeps failing. I cant tell what could cause the issue.

@CharlesChiuGit Would verbose logging provide additional insights?
```sh
cd ~/.local/share/nvimdots/site/lazy/copilot.lua/
git status && GIT_TRACE=1 GIT_CURL_VERBOSE=1 GIT_TRACE_PACKET=1 GIT_TRACE_SETUP=1 GIT_TRACE_SHALLOW=1 git restore --source=HEAD :/
```
I once came across this and eventually realized it was a permission issue. Specifically, the default umask got accidentally changed, leaving that git folder wide open for everyone.
***
By the way can u please help me do a sanity check to ensure that the executables are correctly linked on macOS? I'm just diving into Nix so I might not be much help here. Basically, all u need to do is run `otool -L` on some of the MachO binaries produced by Nix or Mason and confirm whether they're indeed using packages (libraries) provided by Nix rather than those by the system (this ensures that `dyld` is doing the right thing and that packages can be located correctly with our config).

misumisumi:
> By the way can u please help me do a sanity check to ensure that the executables are correctly linked on macOS? I'm just diving into Nix so I might not be much help here. Basically, all u need to do is run otool -L on some of the MachO binaries produced by Nix or Mason and confirm whether they're indeed using packages (libraries) provided by Nix rather than those by the system (this ensures that dyld is doing the right thing and that packages can be located correctly with our config).

Maybe I can incorporate a simple test script into `flake.nix`.
However, I have never written a proper test script, so I need to check the home-manager test scripts, etc.
I'm busy tomorrow, so I'll work on it again after Thursday.

CharlesChiuGit:
> `copilot.lua` and `nvim-nio` keeps failing to checkout no matter what i do, but it works fine outside nix devshell.

omg, it suddenly works w/o any modification. it's so weird lol
image

CharlesChiuGit:
> can u please help me do a sanity check to ensure that the executables are correctly linked on macOS?

ok, let me check. hope i will know how to use those tools soon lol

misumisumi:
Fixed thiree things.
One is put `lazy-lock.json` under `$XDG_CONFIG_HOME/$NVIM_APPNAME`.
Two is rename `test.nix` to `testEnv.nix`.
Three is adding `check-linker` script.
This is simply script to check whether binary installed by mason link to `/nix/store`.
Print cmd name and target lib if not link.

CharlesChiuGit:
seems working exactly like we wanted!
image

CharlesChiuGit:
image

misumisumi:
I have probrem, loop never reaches the end in linux.
Please wait a moment while I fix it.

misumisumi:
It stopped because `set -o pipefail` was set when grep did not match.
Also, some packages could not find libraries because there was no `LIBRARY_PATH` and so on set in `nvim`.
Therefore, I have made it so that the environment variables set in `nvim` are `source`.

flake.nix
```diff
@@ -4,11 +4,98 @@
...
+        };
+        systems = [ "x86_64-linux" "aarch64-linux" "x86_64-darwin" "aarch64-darwin" ];
+        perSystem = { pkgs, system, ... }: {
+          packages = {
+            testEnv = (import ./nixos/testEnv.nix { inherit inputs pkgs; }).activationPackage;
+            check-linker = pkgs.writeShellApplication {
+              name = "check-linker";
+              text = ''
+                #shellcheck disable=SC1090
+                source <(head -n-1 "${self.packages.${system}.testEnv}/home-path/bin/nvim")
```
CharlesChiuGit, line 36: minor typo:
`-n-1` -> `-n 1`
image

misumisumi: Maybe Mac's head doesn't allow negative numbers?
When use negative numbers in GNU ver, output excluding the last n lines.

CharlesChiuGit: aha, then `-n -1`

CharlesChiuGit: image
hmmm, it seems it cant take `-1`...

CharlesChiuGit: `source <(head -n-1 "${self.packages.${system}.testEnv}/home-path/bin/nvim")`

why sourcing last line of this file?

misumisumi: use `sed` instead of `head`.
Can you run cmd ?

CharlesChiuGit: on `aarch64-darwin`:
image

on `aarch64-linux`:
image

misumisumi: Can you show me the output of `ldd $(readlink -f ~/.local/share/nvim/mason/bin/stylua)`?

CharlesChiuGit: ohno, macos doesn't ship with `ldd`, so we need to add this in our pakcages. all results above might be invalid.
image

CharlesChiuGit: using `otool -L` on `aarch64-darwin`:
image

CharlesChiuGit: ldd uses ld to load executable files, and recursively loads dynamically-linked libraries. So using ldd requires being on the target system (e.g., Linux). Thus, ldd cannot be used for ELF files on macOS.

https://stackoverflow.com/a/67818942

misumisumi: Ah. I change using `glibc` in nixpkgs.

CharlesChiuGit: sadly
```
error: Package ‘glibc-nolibgcc-2.39-5’ in /nix/store/qxrwp9qzj2by5vqy4gm6f29njjkw4cna-source/pkgs/development/libraries/glibc/default.nix:184 is not available on the requested hostPlatform:
         hostPlatform.config = "aarch64-apple-darwin"
```

CharlesChiuGit: on `aarch64-linux`:
image

CharlesChiuGit: i think it easier for us to use `otool -L` on darwin and `ldd` on linux.

misumisumi: Maybe, could have fixed it.

misumisumi: > on aarch64-linux:

This may be a problem due to respecting the currently set LIBRARY_PATH and so on.
I think if you remove `nvimdots/mason`, make sure that no environment variables are set, and then install and build again, it will link to.

CharlesChiuGit:
seems like we need to use multiple version of `glibc` on aarch64-linux?
image

flake.nix
CharlesChiuGit: i removed `~/.local/share/nvim/mason`, reinstalled, but still the same.
image

misumisumi: Maybe `stylua` is provided as a binary package.
I think that `mason.nvim` is designed to use the binary if it is provided, and build it if it is not.

CharlesChiuGit: aha, ys. it should be the case.
https://github.com/mason-org/mason-registry/blob/main/packages/stylua/package.yaml

misumisumi:
Can you check node path with `readlink -f $(which node`?
Maybe use system wide node.

CharlesChiuGit:
image
 oh damn, it is using the system node.

CharlesChiuGit:
i disable system node temporarily and enter devshell, now all `node/npm` relate installation works porperly.

misumisumi:
Finally, regarding the github actions that update `flake.lock`, should I update `flake.lock` when updating `lazy-loc.json`, or should I update `flake.lock` separately?
Also, is it better to open a PR before updating, like with dependabot?

CharlesChiuGit:
I prefer update `flake.lock` separately, since I think update it once a month is enough.

And I prefer to update `flake.lock` directly w/o opening a PR.

misumisumi:
I add workflow of auto flake update.
This is called at 2:30 (UTC) of 1st of every month

misumisumi:
I think all the work is completed, what do you think?

CharlesChiuGit:
1. Set check linker uses `otool -L` on `aarch64-darwin` and `ldd` on `linux`.
2. Below error haven't been fixed:
```
error: Package ‘glibc-nolibgcc-2.39-5’ in /nix/store/qxrwp9qzj2by5vqy4gm6f29njjkw4cna-source/pkgs/development/libraries/glibc/default.nix:184 is not available on the requested hostPlatform:
         hostPlatform.config = "aarch64-apple-darwin"
```

CharlesChiuGit:
i'll test it next week or maybe @Jint-lzxy can help lol

CharlesChiuGit (approved):
image
image
looks fine on both `aarch64-darwin` and `aarch64-linux`

Jint-lzxy:
> > can u please help me do a sanity check to ensure that the executables are correctly linked on macOS?
> 
> sure, let me check. hope i will know how to use those tools soon lol

@CharlesChiuGit Apple does indeed have a [thorough documentation](https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/DynamicLibraries/100-Articles/DynamicLibraryUsageGuidelines.html) on dynamic library usage tucked away in its documentation archives (way better than what it offers nowadays lol). It's definitely worth a look if u're into this stuff XD

misumisumi:
@CharlesChiuGit @CharlesChiuGit 
The wiki needs to be updated.
We'll be working on that soon.
We'll report back here when it's done.

PR #1265, opened 2024-05-24
fix: utilize `ts-context` in `glance` preview window

mjkx5:
default zindex of glance [dnlhc/glance.nvim] floatterm is 45. It can display on glance.

### DIFF
--- a/lua/modules/configs/editor/ts-context.lua
+++ b/lua/modules/configs/editor/ts-context.lua
@@ -5,7 +5,7 @@ return function()
 		min_window_height = 0, -- Minimum editor window height to enable context. Values <= 0 mean no limit.
 		line_numbers = true,
 		multiline_threshold = 20, -- Maximum number of lines to collapse for a single context line
-		trim_scope = "outer", -- Which context lines to discard if `max_lines` is exceeded. Choices: 'inner', 'outer'
+		trim_scope = "inner", -- Which context lines to discard if `max_lines` is exceeded. Choices: 'inner', 'outer'
 		mode = "cursor", -- Line used to calculate context. Choices: 'cursor', 'topline'
 		zindex = 50,
 	})
--- a/lua/modules/configs/editor/ts-context.lua
+++ b/lua/modules/configs/editor/ts-context.lua
@@ -7,6 +7,6 @@ return function()
 		multiline_threshold = 20, -- Maximum number of lines to collapse for a single context line
 		trim_scope = "outer", -- Which context lines to discard if `max_lines` is exceeded. Choices: 'inner', 'outer'
 		mode = "cursor", -- Line used to calculate context. Choices: 'cursor', 'topline'
-		zindex = 30,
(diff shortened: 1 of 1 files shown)


CharlesChiuGit:
how to trigger the error?

mjkx5:
zindex 
nvim-treesitter/nvim-treesitter-context : default 20
dnlhc/glance.nvim: default 45, 
vim.api.nvim_open_win({zindex: })

---               • zindex: Stacking order. floats with higher `zindex` go on
---                 top on floats with lower indices. Must be larger than
---                 zero.

CharlesChiuGit:
I'm not asking why, I'm asking how.

I don't see the point of setting `ts-context`'s `zindex` to a higher value.
`Glance` should be on top of other things since it's a temporary thing other than `ts-context`.

mjkx5:
eeeeeeeeeeeeeeee

It is result of that config.

Screenshot_24-May_16-57-17_18805

if not that config, function SetNodeConfig() isn't display.

mjkx5:
for example, I press `gh`, open the preview popup window. If not that config, I need `<c-w>h` into preview, then `[[` jump to position of function name. when look at function name, I realize if that is position that I want to jump to. 
if not, I neet to jump items list , to repeat next. If true, I need to return that position.

use that config, I don't need to jump function name.


In fact, that is still not the best, I need to jump left preview window to display function name. If cursor is right window, It display the function name, which is the best, but I don't have time to do that these two days.

mjkx5:
for example, I look at `mod test` In preview window, I can realize that It isn't that I want. 
Screenshot_25-May_09-14-15_14005

mjkx5:
In my case, `trim_scope` set inner better than outer. Have someone using inner?

CharlesChiuGit:
> for example, I look at mod test In preview window, I can realize that It isn't that I want.

aha, that make sense to me now. tks for the clarification!

CharlesChiuGit (approved):
lgtm

CharlesChiuGit:
> In my case, `trim_scope` set inner better than outer. Have someone using inner?

never tried it, i didn't notice major difference in my use case. but i think it's fine to change it.

mjkx5:
@CharlesChiuGit  sorry, I am not familiar with pull request. I use --force, I accidentally delete your commit. please repeat commit.

mjkx5:
@CharlesChiuGit I am not familiar with pull request. I accidentally used push --force, that delete your commit. 

Please repeat commit, thanks.

mjkx5:
max_lines is set to 3. that means that It can display three lines in window header.

eg: if display text is 4 lines.

mod modName{
   func functionName{
      switch {
          A : 
          B:
      }
   }
}

In that case, set `inner` discard case A. set inner discard modname.

as comment, Which context lines to discard if `max_lines` is exceeded.

ayamir (approved):

ayamir (approved):

PR #1296, opened 2024-06-17
chore(global): cleanup legacy code

Jint-lzxy:
All of these "workarounds" should no longer be necessary since Neovim v0.6 or so.

---

Created with `gh pr create`


CharlesChiuGit (approved):
LGTM

ayamir (approved):

PR #1303, opened 2024-06-20
Feat/telescope collections keymaps

ayamir:
#1299

### DIFF
--- a/lua/modules/configs/ui/alpha.lua
+++ b/lua/modules/configs/ui/alpha.lua
@@ -56,7 +56,7 @@ return function()
 		),
 		button(
 			"space f f",
-			icons.documents.FileFind .. "Telescope file",
+			icons.documents.FileFind .. "Find files",
 			leader,
 			nil,
 			{ noremap = true, silent = true, nowait = true }
--- a/lua/modules/configs/ui/alpha.lua
+++ b/lua/modules/configs/ui/alpha.lua
@@ -62,14 +62,14 @@ return function()
 			{ noremap = true, silent = true, nowait = true }
 		),
 		button(
-			"space f r",
+			"space f d",
 			icons.ui.FolderWithHeart .. "Retrieve dossiers",
 			leader,
 			nil,
 			{ noremap = true, silent = true, nowait = true }
 		),
 		button(
-			"space f w",
+			"space f p",
 			icons.documents.Word .. "Find patterns",
 			leader,
 			nil,
--- a/lua/modules/configs/ui/alpha.lua
+++ b/lua/modules/configs/ui/alpha.lua
@@ -55,20 +55,6 @@ return function()
 			{ noremap = true, silent = true, nowait = true }
 		),
 		button(
-			"space f g",
-			icons.git.Git .. "Telescope git",
-			leader,
-			nil,
-			{ noremap = true, silent = true, nowait = true }
-		),
-		button(
-			"space f w",
-			icons.ui.FolderWithHeart .. "Telescope workspace",
-			leader,
-			nil,
-			{ noremap = true, silent = true, nowait = true }
-		),
-		button(
 			"space f f",
 			icons.documents.FileFind .. "Telescope file",
 			leader,
@@ -76,8 +62,22 @@ return function()
 			{ noremap = true, silent = true, nowait = true }
 		),
 		button(
-			"space f l",
-			icons.documents.Word .. "Telescope live_grep",
+			"space f r",
+			icons.ui.FolderWithHeart .. "Telescope rootdir",
+			leader,
+			nil,
+			{ noremap = true, silent = true, nowait = true }
+		),
+		button(
+			"space f w",
+			icons.documents.Word .. "Telescope word",
+			leader,
+			nil,
+			{ noremap = true, silent = true, nowait = true }
+		),
+		button(
+			"space f g",
+			icons.git.Git .. "Telescope git",
 			leader,
 			nil,
 			{ noremap = true, silent = true, nowait = true }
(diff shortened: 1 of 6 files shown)


CharlesChiuGit:
what(
i dont get it?

CharlesChiuGit (approved):
LGTM

Jint-lzxy (commented):
Shall we also update `dots.tutor` here as well? I think it would be great to introduce this to new users lol

lua/modules/configs/ui/alpha.lua
```diff
@@ -39,55 +39,56 @@ return function()
...
+		button(
+			"space f c",
+			icons.misc.Neovim .. "Telescope collections",
+			leader,
+			nil,
+			{ noremap = true, silent = true, nowait = true }
+		),
+		button(
+			"space f f",
+			icons.documents.FileFind .. "Telescope file",
```
Jint-lzxy, line 59:
```suggestion
			icons.documents.FileFind .. "Find files",
```

imo it's a bit weird to call that "Telescope file," especially for newcomers ~~(or maybe they just wanna find files in the Milky Way, who knows)~~. Perhaps "Find files" would be better?

lua/modules/configs/ui/alpha.lua
```diff
@@ -39,55 +39,56 @@ return function()
...
+			leader,
+			nil,
+			{ noremap = true, silent = true, nowait = true }
+		),
+		button(
+			"space f r",
+			icons.ui.FolderWithHeart .. "Telescope rootdir",
```
Jint-lzxy, line 66:
```suggestion
			icons.ui.FolderWithHeart .. "Retrieve dossiers",
```
Same here. Tho I can't seem to find a better word than `dossier`; any good ideas?

lua/modules/configs/ui/alpha.lua
```diff
@@ -39,55 +39,56 @@ return function()
...
+			leader,
+			nil,
+			{ noremap = true, silent = true, nowait = true }
+		),
+		button(
+			"space f w",
+			icons.documents.Word .. "Telescope word",
```
Jint-lzxy, line 73:
```suggestion
			icons.documents.Word .. "Find patterns",
```
ditto

lua/modules/configs/ui/alpha.lua
```diff
@@ -39,55 +39,56 @@ return function()
...
+			leader,
+			nil,
+			{ noremap = true, silent = true, nowait = true }
+		),
+		button(
+			"space f g",
+			icons.git.Git .. "Telescope git",
```
Jint-lzxy, line 80:
```suggestion
			icons.git.Git .. "Locate Git objects",
```
ditto

lua/modules/configs/ui/alpha.lua
```diff
@@ -39,55 +39,56 @@ return function()
...
+			leader,
+			nil,
+			{ noremap = true, silent = true, nowait = true }
+		),
+		button(
+			"space f m",
+			icons.misc.Ghost .. "Telescope misc",
```
Jint-lzxy, line 87:
```suggestion
			icons.misc.Ghost .. "Miscellaneous artifacts",
```
Errrr I'm running out of ideas here lol

Jint-lzxy (changes requested):

ayamir:
The reason I changed the key bindings was to keep them as consistent with old keymaps as possible.

ayamir:
> Shall we also update `dots.tutor` here as well? I think it would be great to introduce this to new users lol

forgot this, will commit for it later.

Jint-lzxy:
@ayamir I've replaced all instances of "root directory management" with either "(current) working directory management" or "dossier management" bc I really couldn't find any references to the original term. Could u kindly explain why u chose that term, as I might have missed something (quite likely lol)?

ayamir:
> @ayamir I've replaced all instances of "root directory management" with either "(current) working directory management" or "dossier management" bc I really couldn't find any references to the original term. Could u kindly explain why u chose that term, as I might have missed something (quite likely lol)?

It's ok for me. The reason why I use `rootdir` is to keep consistency with the keymap i.e. `<leader>fr`. BTW, `<leader>fw` is better to keep the functionality to find "word" IMO. We should not be able to introduce or introduce new keymaps to prevent custom configuration conflicts with users.

Jint-lzxy:
> We should introduce no or as few new keymaps as possible to prevent conflicts with user customizations.

Yeah, that's true, but I don't think it's ideal for long-term use since the way Telescope works has _fundamentally_ changed. For example, `collection = "word"` isn't really searching for a specific "word" but rather "patterns" in a broader sense, and `collection = "rootdir"` isn't just about projects' root directories but encompasses general dossier management (or a _collection_ of files). For instance, `Zoxide` operates on the current directory, which isn't necessarily the _root_ directory unless we're specifically referring to [chroot(2)](https://man.freebsd.org/chroot(2)) or the current _project's_ root directory. afaiu \*BSD, macOS, and probably Linux (though I'm not as familiar) never use "root directory" in the context of [chdir(2)](https://man.freebsd.org/chdir(2)). So imho it's indeed challenging to rebuild that muscle memory, but creating a mnemonic is far more important for me lol

> The reason why I use `rootdir` is to keep consistency with the keymap i.e. `<leader>fr`.

... hence I would suggest changing `rootdir` to `dossier`, and the corresponding keymap to `<leader>fd`

> BTW, `<leader>fw` is better to keep the functionality to find "word" IMO.

... and use `<leader>fp` for `pattern`.

ayamir:
OK, it's fine to me.

PR #1305, opened 2024-06-21
feat(treesitter): lessened performance restrictions

Jint-lzxy:
This commit lifts some of the performance restrictions previously imposed on Treesitter when incremental parsing was not yet supported. Specifically:

Treesitter will now only be disabled by default for files exceeding 7,500 lines or 2 MiB in size. Testing on an Intel i9-9880H with Apple SSD AP1024N shows noticeable lag at these thresholds.

Further testing on different CPUs and/or SSD combinations is welcome lol

### DIFF
--- a/lua/modules/configs/completion/cmp.lua
+++ b/lua/modules/configs/completion/cmp.lua
@@ -171,7 +171,7 @@ return function()
 				name = "buffer",
 				option = {
 					get_bufnrs = function()
-						return vim.api.nvim_list_bufs()
+						return vim.api.nvim_buf_line_count(0) < 7500 and { vim.api.nvim_list_bufs() } or {}
 					end,
 				},
 			},
--- a/lua/modules/configs/editor/bigfile.lua
+++ b/lua/modules/configs/editor/bigfile.lua
@@ -16,15 +16,16 @@ return function()
 	}
 
 	require("modules.utils").load_plugin("bigfile", {
-		filesize = 1, -- size of the file in MiB
+		filesize = 2, -- size of the file in MiB
 		pattern = { "*" }, -- autocmd pattern
 		features = { -- features to disable
+			"indent_blankline",
 			"lsp",
-			"treesitter",
 			"syntax",
+			"treesitter",
 			"vimopts",
-			ftdetect,
 			cmp,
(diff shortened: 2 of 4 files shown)


lua/modules/configs/completion/cmp.lua
```diff
@@ -171,7 +171,7 @@ return function()
 				name = "buffer",
 				option = {
 					get_bufnrs = function()
-						return vim.api.nvim_list_bufs()
+						return vim.api.nvim_buf_line_count(0) < 7500 and { vim.api.nvim_list_bufs() } or {}
```
CharlesChiuGit, line 174: I got an error whenever i enter `Insert` mode:
圖片

CharlesChiuGit (changes requested):

lua/modules/configs/completion/cmp.lua
Jint-lzxy: > I got an error whenever i enter `Insert` mode

~~So does that mean according to Lua, 0 is not considered a number? 💀~~

nvm I think this should be an upstream issue since `cmp-buffer` doesn't access our config in `buffer.start_indexing_timer` [(where the error occurred)](https://github.com/hrsh7th/cmp-buffer/blob/3022dbc9166796b644a841a02de8dd1cc1d311fa/lua/cmp_buffer/buffer.lua#L134).

ayamir:
For my machine (AMD Ryzen 7 6800HS with Sumsung 980 Pro), there is a noticeable delay (about 2~3s) when trying to enter characters **for the first time** for a cpp file with 7495 lines. The startuptime is 261.81ms. I didn't meet the error above BTW.

ayamir:
These settings are ok for macbook pro 16-inch, 2019 with 2.6 GHz Intel Core i7 (six cores). The startuptime is 316.72ms and no delay when trying to enter characters.

ayamir (approved):

ayamir (changes requested):
Same error when editing a normal size file.
image

ayamir (approved):
LGTM

CharlesChiuGit (approved):
LGTM

PR #1310, opened 2024-06-27
feat(nixos): add mergeLazyLock option

misumisumi:
https://github.com/ayamir/nvimdots/pull/1260#issuecomment-2191929399

`lazy-lock.json` managed by `nix/store` is symbolically linked to `${XDG_CONFIG_DIR}/nvim/lazy-lock.json`, but since it is read-only, there is a problem that it has low affinity with lazy.nvim.
Therefore, lazy-lock.json is managed locally, and when lazy-lock.json in the repository is updated, it is merged to achieve both reproducibility and convenience.

It will be updated with the workflow shown below when home-manager activation.

1. `lazy-lock.json` in the repository is managed as `lazy-lock.fixed.json` in `/nix/store`.
2. When created symlink to `${XDG_CONFIG_DIR}/nvim`, one of the following will be executed.
    1. Not existing `lazy-lock.json` under `${XDG_CONFIG_DIR}/nvim`, copy to then.
    2. Existing `lazy-lock.json` under then, merge two json. At this time, merge to respect the package version of the repository.
    
This workflow will not run unless lazy-lock.json in the repository is updated (path to /nix/store is not updated) (i.e. not every activation of home-manager)

The traditional `bindLazyLock` options are also retained,

### DIFF
--- a/nixos/neovim/default.nix
+++ b/nixos/neovim/default.nix
@@ -8,9 +8,10 @@ let
   cfg = config.programs.neovim.nvimdots;
   inherit (lib) flip warn const;
   inherit (lib.attrsets) optionalAttrs;
-  inherit (lib.options) mkEnableOption mkOption mkIf literalExpression;
-  inherit (lib.strings) concatStringsSep versionOlder versionAtLeast;
   inherit (lib.lists) optionals;
+  inherit (lib.modules) mkIf;
+  inherit (lib.options) mkEnableOption mkOption literalExpression;
+  inherit (lib.strings) concatStringsSep versionOlder versionAtLeast;
   inherit (lib.types) listOf coercedTo package functionTo;
 in
 {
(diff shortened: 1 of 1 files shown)


CharlesChiuGit (approved):
LGTM, `lazy-lock.json` is not read-only.
image

misumisumi:
> This workflow will not run unless lazy-lock.json in the repository is updated (path to /nix/store is not updated) (i.e. not every activation of home-manager)

Is this behavior okay? or do we try to merge every time you activate home-manager?

misumisumi:
Do I merge this ?

nixos/neovim/default.nix
```diff
@@ -19,6 +26,13 @@ in
           Bind lazy-lock.json in your repository to $XDG_CONFIG_HOME/nvim.
           Very powerful in terms of keeping the environment consistent, but has the following side effects.
           You cannot update it even if you run the Lazy command, because it binds read-only.
+          You need to remove lazy-lock.json before activation, if `mergeLazyLock` have set.
```
Jint-lzxy, line 29:
```suggestion
          You need to remove lazy-lock.json before enabling this option if `mergeLazyLock` is set.
```

nixos/neovim/default.nix
```diff
@@ -19,6 +26,13 @@ in
...
+        '';
+        mergeLazyLock = mkEnableOption ''
+          Merge into already existing lazy-lock.json in $XDG_CONFIG_HOME/nvim every activation.
+          This will respect the package version of lazy-lock.json in the repository.
+          Achieve environmental consistency while being flexible to change.
+          You need to unlink lazy-lock.json before activation, if `bindLazyLock` have set.
```
Jint-lzxy, line 35:
```suggestion
          Merges the managed lazy-lock.json with the existing one under $XDG_CONFIG_HOME/nvim upon each activation.
          This means changes to lazy-lock.json in the config directory (likely due to updates) will be preserved.
          In other words, it achieves environment consistency while remaining adaptable to changes.
          You need to unlink lazy-lock.json before enabling this option if `bindLazyLock` is set.
```

Jint-lzxy (commented):
Apologies for the delay! Here are some stylistic suggestions, feel free to use any if u think they're okay lol

nixos/neovim/default.nix
misumisumi: Strictly speaking, changes to the lazy-lock.json config file are not always retained.
However, since this explanation would be too long, I have added a note to ask users to refer to the wiki for details on which files have higher priority and their behavior.
I will write more details on the wiki once the changes have been merged.

Jint-lzxy (approved):
LGTM

PR #1323, opened 2024-07-13
feat(which-key): `v3` support

channinghsu:
This pull request updates the which-key mappings to the new specification as outlined in the documentation for [version 3](https://github.com/folke/which-key.nvim?tab=readme-ov-file#%EF%B8%8F-configuration:~:text=a%20regular%20keymap.-,Warning,use%20the%20new%20add%20method%20if%20you%20updated%20your%20existing%20mappings.,-Mappings%20can%20be). The previous mappings were using the old specification, which caused warnings during the health check. The updated mappings now use the add method and are structured according to the latest guidelines.
## Before:
5701720888312_ pic
5711720888477_ pic
## After:
image

### DIFF
--- a/lua/modules/configs/tool/which-key.lua
+++ b/lua/modules/configs/tool/which-key.lua
@@ -2,43 +2,19 @@ return function()
 	local icons = {
 		ui = require("modules.utils.icons").get("ui"),
 		misc = require("modules.utils.icons").get("misc"),
-		git = require("modules.utils.icons").get("git", true),
-		cmp = require("modules.utils.icons").get("cmp", true),
 	}
 
-	require("which-key").register({
-		["<leader>"] = {
-			b = {
-				name = icons.ui.Buffer .. " Buffer",
-			},
-			d = {
-				name = icons.ui.Bug .. " Debug",
-			},
-			f = {
-				name = icons.ui.Telescope .. " Fuzzy Find",
-			},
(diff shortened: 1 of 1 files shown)


CharlesChiuGit (approved):
wow, so fast!
LGTM

圖片
some icons are missing, but i think it's acceptable.

Jint-lzxy (commented):
LGTM! And u r suuuper fast lol But I kinda wanna address a few issues with regards to the new version, so I've opened [another version](https://github.com/ayamir/nvimdots/pull/1324) that does the exact same thing but with some settings explicitly disabled, specifically:
- Per https://github.com/folke/which-key.nvim/issues/656#issuecomment-2227098307, I disabled `which-key` in all modes other than normal and insert mode bc imo having that popup appear every time I'm about to make a selection is _remarkably_ annoying and a major distraction.
- Disabled the built-in icon support. afaiu folke seems to have used some pattern-matching magic to detect which icon to use, but to be blunt it's buggy and doesn't cover all cases (my observations). Plus, imho it's really just another layer of distraction since we've already assigned an intuitive icon to each top-level definition.

ayamir:
Personally I prefer keep consistency with the old appearance. What's more, this PR seems also missing on the icons like this:
image
So I merged #1324, thanks for your contribution again.

Issue #1339, opened 2024-08-12
rtp plugins, aka distribution plugins does not actually got disabled


CharlesChiuGit:
### Version confirmation

- [X] Confirm

### Following prerequisites

- [X] Confirm

### Not a user config issue

- [X] Confirm

### Neovim version

NVIM v0.10.1

### Operating system/version

macOS 14.6.1

### Terminal name/version

wezterm

### Branch info

main (Default/Latest)

### Fetch Preferences

SSH (use_ssh = true)

### How to reproduce the issue

1.  `:Lazy profile`
2. Some rtp plugins is still listed, even we disable it via `disable_distribution_plugins()`.
![image](https://github.com/user-attachments/assets/9614cb59-d01b-4400-be21-be194feca00b)

### Expected behavior

those rtp plugins don't load at all.

CharlesChiuGit:
```lua
		performance = {
				...
				disabled_plugins = {
					"gzip",
					"matchit",
					"matchparen",
					"netrwPlugin",
					"tarPlugin",
					"tohtml",
					"tutor",
					"zipPlugin",
					"rplugin",
					"spellfile",
				},
			},
		},
```

set those rtp pluigins in `lazy` can disable rtp plugins.
I will open a pr to fix this.

PR #1340, opened 2024-08-12
fix(core): disable rtp plugins via `lazy.nvim`, #1339

CharlesChiuGit:
as showed in #1339 

from
image
 to
image


ayamir:
we can reserve the comments to note.

ayamir (approved):

PR #1341, opened 2024-08-13
fix: some vim plugins get disabled by mistake

Jint-lzxy:
This commit addresses a few regressions introduced by #1340:

* The `tutor` plugin was accidentally disabled.
* netrw, along with its configuration, was mistakenly disabled.
* EditorConfig is now being implicitly disabled, which wasn't intended before #1340.

### DIFF
--- a/lua/core/init.lua
+++ b/lua/core/init.lua
@@ -22,14 +22,6 @@ local createdir = function()
 	end
 end
 
-local load_distribution_plugins = function()
-	-- Set this to 0 in order to disable native EditorConfig support
-	vim.g.editorconfig = 1
-
-	-- Newtrw liststyle: https://medium.com/usevim/the-netrw-style-options-3ebe91d42456
-	vim.g.netrw_liststyle = 3
-end
-
 local leader_map = function()
 	vim.g.mapleader = " "
 	-- NOTE:
@@ -104,7 +96,6 @@ end
 
 local load_core = function()
 	createdir()
-	load_distribution_plugins()
 	leader_map()
 
 	gui_config()
--- a/lua/core/init.lua
+++ b/lua/core/init.lua
@@ -22,6 +22,16 @@ local createdir = function()
 	end
 end
 
+local load_distribution_plugins = function()
+	-- Set this to true in order to enable native EditorConfig support
+	-- WARN: Sleuth.vim already includes all the features provided by this plugin.
+	--       Do NOT enable both at the same time, or you risk breaking the entire detection system.
+	vim.g.editorconfig = false
+
+	-- Newtrw liststyle: https://medium.com/usevim/the-netrw-style-options-3ebe91d42456
+	vim.g.netrw_liststyle = 3
+end
+
 local leader_map = function()
 	vim.g.mapleader = " "
 	-- NOTE:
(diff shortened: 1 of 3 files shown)


CharlesChiuGit:
EditorConfig is not disabled currently.
image

CharlesChiuGit:
i thought we use `nvim-tree` as netrw, no?
or it's necessary to load builtin netrw?

ayamir (approved):

ayamir:
I don't use editorconfig, so I can't judge if it works for fine.
And as the readme of `nvim-tree` said, disabling `netrw` is strongly advised.

Jint-lzxy:
Sorry for the late reply! I've had a lot to sort through over the past few days lol

> EditorConfig is not disabled currently

Oops, that was enabled by mistake. Sorry for the extra noise introduced with this PR! I somehow forgot we already have [sleuth.vim](https://github.com/tpope/vim-sleuth) lmao

> i thought we use `nvim-tree` as netrw, no?
> or it's necessary to load builtin netrw?

`nvim-tree` isn't actually hijacking `netrw` during startup cuz it's [intended to be loaded on demand](https://github.com/ayamir/nvimdots/blob/8d81c66efe75604e8ec7de4238f72122da0a3065/lua/modules/plugins/tool.lua#L22-L32). So if we disable `netrw`, users might encounter issues when trying to open directories with neovim _at startup_. Alternatively, we could configure `nvim-tree` to hijack `netrw`, but that would likely cause a noticeable increase in startup time[^1].

> And as the readme of `nvim-tree` said, disabling `netrw` is strongly advised.

Hmm, I _think_ this ultimately comes down to personal preference cuz afaiu `netrw` is generally useful for opening a directory with neovim (like `nvim .`) at startup, which is significantly faster than using `nvim-tree`[^1] and doesn't come with a huge deficit in features compared to `nvim-tree`. In most other cases, `nvim-tree` handles things well.

[^1]: That's what happens on my devices, tho I'm not sure if it's different with other software and hardware combos.

Jint-lzxy:
https://github.com/ayamir/nvimdots/pull/1341/commits/0b6dc571dbde61f4d1f2e1c3e7c50e2f30aaf7ec reintroduces `load_distribution_plugins` and adds a warning about enabling `editorconfig.lua`:
```console
Sleuth.vim already includes all the features provided by this plugin.
Do NOT enable both at the same time, or you risk breaking the entire detection system.
```
I've kept `netrw` enabled for now, but I'm open to changing that whenever lol

ayamir:
Fine for me.

aarnphm:
tested with builtin editorconfig seems to work (as I don't use sleuth.vim).

CharlesChiuGit (approved):
LGTM

CharlesChiuGit:
we can't disable builtin `editorconfig` via global options, if we use `lazy.nvim`.
we can only get a valid disable via `lazy.nvim`.

CharlesChiuGit:
tbh, i think don't we need `vim-sleuth` since we now have a builtin `editorconfig`.
how do u guys think? @ayamir @Jint-lzxy @aarnphm

aarnphm:
sgtm. always +1 when removing plugins.

ayamir:
> tbh, i think we don't need `vim-sleuth` since we now have a builtin `editorconfig`.

`vim-sleuth` is useful for me b/c I don't use `editorconfig` lol

CharlesChiuGit:
ok lol
then i think it's ready to merge.

PR #1351, opened 2024-09-17
feat(lazygit): add lazygit

vkumbhar94:


CharlesChiuGit:
we already have lazygit keymaps
https://github.com/ayamir/nvimdots/blob/6c1933580ed297cdcfc2094abb815cbb1473e996/lua/keymap/tool.lua#L58-L63

vkumbhar94:
Thanks @CharlesChiuGit my bad, i didn't notice it.

PR #1361, opened 2024-10-27
feat(plugins)!: Remove neodim

Jint-lzxy:
This commit removes neodim because it's been two years since semantic tokens were introduced in LSP Specification 3.16.0, and it seems like most servers have adopted all those features by now. Plus, this plugin has been buggy at times as noted in previous issues, and it can be a hassle to troubleshoot.


ayamir (approved):
Agree, I have been disabled it in my own config for a long times.

CharlesChiuGit (approved):
lgtm

PR #1362, opened 2024-11-02
Delete duplicate keymaps

AngelontheRoad:
These keymaps are redefined to smart-splits.nvim in keymap/editor.lua

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -5,11 +5,11 @@
         "nixpkgs": "nixpkgs"
       },
       "locked": {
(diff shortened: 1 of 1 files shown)


AngelontheRoad:
Is it possible to put windows manager keymaps in a file? Keymaps about create, delete, move, resize or toggle focus of windows/buffers/tabs moves to `wm.lua` which is placed in `lua/keymap`, and leave `mapping.lua` basic keymaps.

PR #1364, opened 2024-11-06
Add keymaps to toggle lsp inlay hints and virtual text

AngelontheRoad:
I found that we have to preset the value of `lsp_inlayhint` and `diagnostics_virtual_text` in `setting.lua` but in fact sometimes we need to toggle their display. So I make this pr and it works well in my PC.


ayamir (approved):
Sorry for the late reply! Recently I'm busy for my autumn recruitment and graduation project mid-term report.

This PR is useful for me. LGTM :+1:

AngelontheRoad:
Good luck finding a good job! I'm sorry that the description is incorrect because they take effect globally not in current buffer. I wonder if it needs to have buffer-toggled keymaps. I would like to create a pr if you do think so or just fix the description.

Issue #1367, opened 2024-11-10
:sparkles: Automatically compile files before debugging (C++)


CodingOIer:
### Feature description

Automatically compile files before debugging, which can save a lot of time

ayamir:
Our debug functionality is impl by [nvim-dap](https://github.com/mfussenegger/nvim-dap) and when you press `<F6>` is invoke `require("dap").continue()` in fact so it will introduce unnecessary complexity for us if we want to impl it.

CodingOIer:
thx

PR #1368, opened 2024-11-17
feat: auto compile before debug (#1367).

ayamir:
#1367


ayamir:
@CodingOIer

CharlesChiuGit:
seems reasonable for me, but i'm currently busy so i can't test it.

CodingOIer:
thx a lot

CodingOIer:
However, the default output path of Debug should also be changed to ./filename.out. Now the default is ./a.out, and the compiled binary file is ./filename.out

ayamir:
> However, the default output path of Debug should also be changed to ./filename.out. Now the default is ./a.out, and the compiled binary file is ./filename.out

Done

CodingOIer:
It works great, thanks a lot

PR #1371, opened 2024-11-29
fix: supports getting of user-defined color palettes

misumisumi:
Due to the change in a631be214b3b537d7b5d7c999e21ba59a3a467fc, an error will now occur if a color scheme other than catppuccin is used.
Previously, when using another color scheme, it seemed to fall back to the defined palette.
This change supports retrieval of user-defined color palettes, allowing you to use a wider range of color schemes.

- error message
image

### DIFF
--- a/lua/modules/utils/init.lua
+++ b/lua/modules/utils/init.lua
@@ -56,39 +56,41 @@ local function init_palette()
 		})
 	end
 
+	local fallback_palette = {
+		rosewater = "#DC8A78",
+		flamingo = "#DD7878",
+		mauve = "#CBA6F7",
+		pink = "#F5C2E7",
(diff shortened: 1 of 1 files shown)


ayamir:
My fault, I didn't test the case when colorscheme is not `catppuccin`.
But the current impl doesn't seem to solve the problem:
image
And the introduction of `get_palette_func` seems to be duplicated with `palette_overwrite`?

ayamir:
This will solve the problem:
image

image

misumisumi:
Oh, I overlooked `palette_overwrite`.
I removed it and changed code.

ayamir (approved):
LGTM

PR #1374, opened 2024-12-06
fix: client filter argument typo

AngelontheRoad:
doc [here](https://neovim.io/doc/user/lsp.html#vim.lsp.get_clients())


ayamir (approved):
LGTM

PR #1381, opened 2024-12-12
fix(utils): hl-Statusline retrieval issue in toggleterm

Jint-lzxy:
This commit should fix a weird bug where `nvim_get_hl` would return an empty dict when trying to get `hl-Statusline` inside `toggleterm`. Not 100% sure why this was happening or what exact conditions caused it, but with this fix lualine now shows the right backgrounds in terminal mode!
- **Before (notice that eg the git section doesn't have proper `guibg`):**
before
- **After:**
after


ayamir (approved):
Not sure its influence but LGTM.

PR #1383, opened 2024-12-29
fix: avoid bufferline not appear when open file by telescope or nvim-tree.

ayamir:
Currently we set `bufferline.nvim` loaded when `BufReadPost`, it will cause the bufferline doesn't appear when we open file by telescope or nvim-tree. This patch make it load when `BufReadPre`.


CharlesChiuGit (approved):
make sense to me.

PR #1390, opened 2025-01-07
fix: search selection pattern under visual mode.

ayamir:
This patch fix the bug that searching selection region doesn't works as expect. Current main branch will search word under cursor rather than selection region.


CharlesChiuGit (approved):
it works, LGTM

PR #1393, opened 2025-01-09
feat: use edgy.nvim to define window layout.

ayamir:
This patch add edgy.nvim to define window layout for better workflow. WIP and need more voice @greggh @CharlesChiuGit


ayamir:
Current configuration with `neo-tree` and screenshot:
image

greggh:
That is exactly what I was hoping for. This is amazing.

ayamir:
Newest configuration with neotree outline and make vertical toggleterm compatible. Seems the outline from aerial is not needed now.
image

ayamir:
> Do the other Trouble windows like definitions and so on, all open on the right as well, under the Symbols?

Just found out that trouble already supports the display of lsp functionalities, I will try to arrange them as reasonably as possible.

ayamir:
latest config:
image

greggh:
That is awesome.

ayamir:
It looks like this now (with `nvim-tree`):

image

ayamir:
@greggh Have you tried this？

greggh:
> @greggh Have you tried this？

Not yet, ill have time to try it later today. Was a super busy weekend. Thanks for all the effort here, it looks great. I'll respond later after I get a chance to play with it.

greggh:
This is great. It is doing exactly what I hoped it would. Testing it out I found myself hoping that the key bind "gt" would bring up the diagnostics on the bottom in the same way as other things. Also I would love it if the quickfix list went down to the bottom just like the toggleterm and help properly.

The "go" key bind does what I wanted and works great with the LSPs I am working with.

ayamir:
> I would love it if the quickfix list went down to the bottom just like the toggleterm and help properly.

It has already configured. Do you mean `Trouble project_diagnostics` window?

ayamir:
> It would also be great to have one key bind that is sort of the focus / full-ide key bind. So when I press those keys it does exactly what control-n, gt, and go all do at once. It would take me from an open buffer that is my full screen to my "full ide" in one key bind.

You can define it on your own by `map_callback` with function. Personally, I don't prefer to define more keymaps for `nvimdots`.

greggh:
> > I would love it if the quickfix list went down to the bottom just like the toggleterm and help properly.
> 
> It has already configured. Do you mean `Trouble project_diagnostics` window?

You're right. For some reason it got messed up on my end where the quick fix list was coming up as half the screen instead of just at the bottom. But I think that was my fault now. Just tested again and its working fine.

PR #1398, opened 2025-01-18
Revert "feat: auto compile before debug (#1367)."

Jint-lzxy:
This PR reverts commit a1c76b23aa347e8fce388c644a44e9f73fe5be44.

The main rationale behind this is IMO the current implementation of this function has several limitations that need to be addressed:

1. It works with only a single file, which is not typical for most use cases
2. It only supports gcc as the compiler, whereas other compilers (most notably clang, xlc, cl.exe and friends) are not supported
3. There's no way to control compile and/or build flags
4. File type detection is too restrictive and doesn't cover all cases (using vim's ftplugins may be more effective)
and
5. Violating any of these assumptions results in errors, rendering this keymap completely ineffective


ayamir:
Tend to improve it rather than eliminate it b/c this will break compatiable, especially this is an enhancement made in response to a issue: https://github.com/ayamir/nvimdots/issues/1367.

Jint-lzxy:
> Tend to improve it rather than eliminate it b/c this will break compatiable.

Agreed, but I think it might be better to move this to eg the FAQ instead of keeping it in the base config...? cuz imo there's just too much to consider if we're actually planning to support it. Even if we managed to make it work, it would be strenuous to maintain. Maybe we could even turn it into a plugin and delegate the work there?

CharlesChiuGit:
https://github.com/stevearc/overseer.nvim

ayamir:
> > Tend to improve it rather than eliminate it b/c this will break compatiable.
> 
> Agreed, but I think it might be better to move this to eg the FAQ instead of keeping it in the base config...? cuz imo there's just too much to consider if we're actually planning to support it. Even if we managed to make it work, it would be strenuous to maintain. Maybe we could even turn it into a plugin and delegate the work there?

Move it to FAQ is acceptable for me.

ayamir:
DONE: https://github.com/ayamir/nvimdots/wiki/Knowledge-Garden#-dap-related-issues

PR #1404, opened 2025-01-19
Revert "fix: support overwrite when merge config. (#1394)"

Jint-lzxy:
This reverts commit 7056741dd38901ff5b615ed2b604e088691255b5.

Sorry for the noise but I'm a lil confused about this new option...? Like idk if the performance difference between these two is even that big, and maybe it's not actually working as expected?

### DIFF
--- a/lua/modules/configs/ui/edgy.lua
+++ b/lua/modules/configs/ui/edgy.lua
@@ -91,5 +91,5 @@ return function()
 				filter = trouble_filter("right"),
 			},
 		},
-	}, false, nil, true)
+	})
 end
--- a/lua/modules/utils/init.lua
+++ b/lua/modules/utils/init.lua
@@ -297,8 +297,7 @@ end
 ---@param opts nil|table @The default config to be merged with
 ---@param vim_plugin? boolean @If this plugin is written in vimscript or not
 ---@param setup_callback? function @Add new callback if the plugin needs unusual setup function
----@param overwrite? boolean @If load user table-type config by overwriting
-function M.load_plugin(plugin_name, opts, vim_plugin, setup_callback, overwrite)
+function M.load_plugin(plugin_name, opts, vim_plugin, setup_callback)
 	vim_plugin = vim_plugin or false
 
 	-- Get the file name of the default config
@@ -331,11 +330,7 @@ function M.load_plugin(plugin_name, opts, vim_plugin, setup_callback, overwrite)
 			if ok then
 				-- Extend base config if the returned user config is a table
 				if type(user_config) == "table" then
-					if overwrite == true then
-						opts = vim.tbl_deep_extend("force", opts, user_config)
-					else
-						opts = tbl_recursive_merge(opts, user_config)
-					end
+					opts = tbl_recursive_merge(opts, user_config)
 					setup_callback(opts)
 				-- Replace base config if the returned user config is a function
 				elseif type(user_config) == "function" then


ayamir:
`overwrite` option is the only solution to delete pre-defined something, like I want to define `edgy.nvim` left region with `NeoTree` rather than `NvimTree`. Otherwise I need to use `function` to repeat lots of content the same with defaults.

Jint-lzxy:
> `overwrite` option is the only solution to delete pre-defined something

But I think this might break some users' expectations of the loader right 🤔 Also think I mentioned `vim.NIL` elsewhere too, it's like a user object so maybe we can use that to signal "erasure" of the field instead?

ayamir:
Marking a option as nil is not enough, overwrite is an irreplaceable way to acheive delete and re-define.

Jint-lzxy:
> Marking a option as nil is not enough, overwrite is an irreplaceable way to acheive delete and re-define.

😭 srry if I wasn't super clear earlier! What I meant was we could let users pass `vim.NIL` inside the override table. Consider `:h vim.NIL`:
```help
vim.NIL                                                              *vim.NIL*
    Special value representing NIL in |RPC| and |v:null| in Vimscript
    conversion, and similar cases. Lua `nil` cannot be used as part of a Lua
    table representing a Dictionary or Array, because it is treated as
    missing: `{"foo", nil}` is the same as `{"foo"}`.
```

To break it down, imagine the original table looks like this:
```lua
local orig = {
	top = {
		middle = {
			opt = "OKAY SOME OPTIONS",
			whatever = "BLA"
		}
	}
}
```

And the user's override is like this:
```lua
local override = {
	top = {
		middle = vim.NIL,
	},
}
```

Now, when we call our `tbl_recursive_merge`, here's what Neovim gives back:
```lua
{
  top = { 
    middle = vim.NIL 
  }
}
```

Then all we need to do is strip out the `vim.NIL` values in the recursive merge function and we're good to go! Also I've seen this (along with `v:null`) in tons of plugins, so I'd say it's a pretty "standard" approach and way more systematic. On top of that the main reason why I want to revert this first is "what's the point of setting `overwrite=true` just for this plugin?" I feel like our users' mileage might vary (possibly by a lot) otherwise.

ayamir:
> 😭 srry if I wasn't super clear earlier! What I meant was we could let users pass vim.NIL inside the override table.

My real motivation for introducing this change is **NOT** to make a part of default configuration be nil, but to use the user's configuration to **replace the part of default configuration**. For this example, `left` option is exclusive for different configuration (nvim-tree or other file tree like neo-tree). If the default configuration for `nvim-tree` is not erased, the left part of edgy will show two file tree like this:

image

> On top of that the main reason why I want to revert this first is "what's the point of setting overwrite=true just for this plugin?" I feel like our users' mileage might vary (possibly by a lot) otherwise.

The reason why I introduce it by adding a parameter to `tbl_recursive_merge` is whether  `overwrite` or not is clearly for one plugin like the layout option of `edgy.nvim`.

Of course, the replace function can be impl by using return a function to achieve this, but it needs lots of redundant code.

ayamir:
This option is useful for `search.nvim` too.

Jint-lzxy:
> My real motivation for introducing this change is NOT to make a part of default configuration be nil, but to use the user's configuration to replace the part of default configuration.

oh lol my bad I think I got what u meant now, srry for the confusion! I think [passing a function as the value for that key](https://github.com/ayamir/nvimdots/blob/d58bf9a0efdf5007425aae438a43850c1b863649/lua/modules/utils/init.lua#L294-L295) should do the trick right? since it's our first branch it should def be respected

ayamir:
> I think [passing a function as the value for that key](https://github.com/ayamir/nvimdots/blob/d58bf9a0efdf5007425aae438a43850c1b863649/lua/modules/utils/init.lua#L294-L295) should do the trick right? since it's our first branch it should def be respected

It's OK.

PR #1418, opened 2025-02-21
fix: update gopls config.

ayamir:
ref: https://github.com/golang/tools/blob/33f1ed9242128736ca381ce86d10a5fc479aab4c/gopls/internal/settings/settings.go#L188


CharlesChiuGit (approved):
lgtm

PR #1420, opened 2025-02-26
fix：修复local-highlight.nvim更新后缺少依赖异常

ttbug:
local-highlight.nvim插件新增了依赖snacks.nvim，导致异常报错 https://github.com/tzachar/local-highlight.nvim/commit/b29692c1edbf292fb7f62552c54bc7b06e61466b


CharlesChiuGit:
i don't think we need to add `snacks.nvim` as dependency?
the animation part doesn't seem necessary.

ttbug:
> i don't think we need to add `snacks.nvim` as dependency? the animation part doesn't seem necessary.

嗯，nvim 0.10以上的版本好像是默认开启的，需要这个依赖。不添加依赖的话，更新一下local-highlight.nvim的配置关掉这个功能也可以

PR #1428, opened 2025-03-03
chore(configs): remove useless legacy configs

Jint-lzxy:
This commit got rid of some legacy configs that literally did nothing (fr this time lol).

### DIFF
--- a/lua/modules/configs/editor/grug-far.lua
+++ b/lua/modules/configs/editor/grug-far.lua
@@ -1,5 +1,4 @@
 return function()
-	vim.g.maplocalleader = ","
 	require("modules.utils").load_plugin("grug-far", {
 		engine = "ripgrep",
 		engines = {
--- a/lua/modules/configs/editor/grug-far.lua
+++ b/lua/modules/configs/editor/grug-far.lua
@@ -1,5 +1,8 @@
 return function()
 	require("modules.utils").load_plugin("grug-far", {
+		windowCreationCommand = "bot split",
+		disableBufferLineNumbers = false,
+		icons = { enabled = true },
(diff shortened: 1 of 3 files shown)


ayamir (approved):

lua/modules/configs/editor/grug-far.lua
```diff
@@ -1,5 +1,4 @@
 return function()
-	vim.g.maplocalleader = ","
```
CharlesChiuGit, line 2: `grug-far` explicitly need to set localleader for it's keymap to work.

CharlesChiuGit (changes requested):

lua/modules/configs/editor/grug-far.lua
Jint-lzxy: I just checked [the source code](https://github.com/MagicDuck/grug-far.nvim/blob/f86a07ee814d9eff3c08a7ce1041e502c32872c5/lua/grug-far/opts.lua#L163-L191) (correct me if I'm wrong!) and looks like `grug-far`'s _default_ is using `<localleader>`, but tbh I think that's not rlly needed...? Since our `<leader>` is diff from `<localleader>`, maybe we should just pull that table out and define it ourselves 🤔

CharlesChiuGit: > maybe we should just pull that table out and define it ourselves 🤔
that's also doable, but since the `grug-far` panel also shows keymap hints.
i'm not sure if the keymap hints will still exist once we define the keymap elsewhere.

ayamir (approved):
LGTM

PR #1429, opened 2025-03-03
chore(matchup): enable some vim-matchup features

Jint-lzxy:
This commit turned on a few vim-matchup features that pair nicely with treesitter imo.


CharlesChiuGit (approved):
lgtm

PR #1430, opened 2025-03-03
feat(dropbar): replace `path` with custom source

Jint-lzxy:
This commit replaces the default `path` source with a custom one to prevent the path section from always displaying (dropbar would incorrectly display the full path when switching eg to a terminal window as it would assume the terminal's cwd as the current working directory)


ayamir (approved):
This patch will cause dropbar showing the filename as the top level.

ayamir:
> This patch will cause dropbar showing the filename as the top level.

This will cause inconvenience when open library file, make it hard to distinguish it with repo files.

PR #1431, opened 2025-03-03
fix(mason): add nil check to on_stderr callback in install-pylsp-plugins

Jint-lzxy:
The luv doc says the `on_stderr` callback is also called whenever the program exits, not just when smth is written to stderr. So we should only notify if there's a real error in `msg_stream`.


CharlesChiuGit (approved):
lgtm

CharlesChiuGit:
@misumisumi could u check if u got time?

misumisumi:
works for my env too!

PR #1435, opened 2025-03-17
Revert "feat(dropbar): replace `path` with custom source"

ayamir:
This patch revert #1430. This will cause inconvenience when open library file, make it hard to distinguish it with repo files.

### DIFF
--- a/lua/modules/configs/tool/dropbar.lua
+++ b/lua/modules/configs/tool/dropbar.lua
@@ -5,43 +5,20 @@ return function()
 		misc = require("modules.utils.icons").get("misc", true),
 		ui = require("modules.utils.icons").get("ui", true),
 	}
-	local utils = require("dropbar.utils")
-	local sources = require("dropbar.sources")
-	-- Custom source to display only the leaf filename in the dropbar
-	sources.symbols = {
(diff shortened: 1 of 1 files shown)


Jint-lzxy:
My bad for the slow reply! Honestly I feel like this should be how it works lol I switched the `path` source to be relative to cwd cuz bufferline alr shows the full path if there's duplicate file names + we also have the cwd in the bottom right (kinda same as how it was before dropbar came in too). Not sure what the issue is here tbh but lmk if I'm missing something!

ayamir:
> we also have the cwd in the bottom right

It's fine, maybe I should change my habbit.

Jint-lzxy:
> It's fine, maybe I should change my habbit.

lol no worries we can def discuss diff options here! my main motivation behind this is just making things more consistent w the other changes we've made but like always we can tweak our own config however we want 🤣 a better way to do this is prob to integrate w `project.nvim` and use the default `path` source with the root set to whatever `project.nvim` reports, but we'd have to turn off `manual_mode` for project.nvim tho... that way all the files share a common root but still stay separate based on the project, and anything outside will just show the full path

PR #1436, opened 2025-03-18
feat(bigfile): use faster.nvim to replace bigfile.nvim.

ayamir:
This patch replace `bigfile.nvim` with `faster.nvim` with faster bigfile load speed. Tested for auto generated rpc interface file with 40000 lines, the first loading time was 2 times faster, and the subsequent jump time was nearly 3 times faster. Because the test was conducted on company private codes, it is impossible to record the screen to show a intuitive comparison. You guys can also test it by yourself to feel the speed up.


CharlesChiuGit:
https://gist.githubusercontent.com/gruehle/2646280/raw/e80c4c9135070592ffdd4a14d7a0e314053f7040/longfile.txt

CharlesChiuGit (approved):
lgtm

PR #1442, opened 2025-03-28
feat(treesitter): remove performance limitations

Jint-lzxy:
This commit removes all previous performance restrictions as treesitter now supports full asynchronous parsing.


ayamir (approved):
LGTM

PR #1443, opened 2025-03-28
fix(lualine): resolve background color defaulting issue

Jint-lzxy:
This commit addresses the root cause of #1381 by allowing lualine to automatically select the default background color. Previously, lualine attempted to retrieve the background manually, which could lead to issues when querying in a different namespace (and resulting in an empty dictionary).


ayamir (approved):
LGTM, I find the colorscheme of bufferline is also won't change when switching to a new colorscheme.

PR #1444, opened 2025-03-28
fix(keymap): prevent new builtins from interfering with our keymap

Jint-lzxy:
This commit prevents lag when using `gr` by avoiding conflicts with new keymaps Neovim defines with this prefix.


ayamir (approved):
LGTM.

PR #1450, opened 2025-04-12
fix: replace project.nvim with newest apis.

ayamir:
The original plugin author hasn't updated this plugin for two years. So I tend to fix it by myself.


CharlesChiuGit (approved):
lgtm

PR #1452, opened 2025-04-17
feat: use tiny inline diagnostics to manage virtual text.

ayamir:
This PR remove the setting entry for virtual text b/c [tiny-inline-diagnostic.nvim](https://github.com/rachartier/tiny-inline-diagnostic.nvim) will manage it.

Effect:
Only show diagnostic message at current cursor position to avoid too many messages disrupt attention. We have `Trouble` and `g[`/`g]` to navigate between multiple diagnostic messages, so this is acceptable for me.

image


CharlesChiuGit (approved):
LGTM, this looks nice!

PR #1456, opened 2025-04-24
refactor: reintro aerial as symbol outline.

ayamir:
Reason: https://github.com/ayamir/nvimdots/discussions/494#discussioncomment-12925108

This pr a is partially revert of #1393, `Trouble lsp` is removed due to we ofter use `gh` to check lsp reference, this window is not helpful enough.

### DIFF
--- a/lua/keymap/completion.lua
+++ b/lua/keymap/completion.lua
@@ -27,6 +27,12 @@ function M.lsp(buf)
 			:with_silent()
 			:with_buffer(buf)
 			:with_desc("lsp: Toggle outline"),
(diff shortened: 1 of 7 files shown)


CharlesChiuGit:
@ayamir sorry, i forget to review this pr. do u still want to add `aerial` back?

ayamir:
> @ayamir sorry, i forget to review this pr. do u still want to add `aerial` back?

I'm testing a new solution to open outline with dynamic width considering the width of rightmost window in the split view. The telescope function seems having support by trouble too.

PR #1458, opened 2025-04-29
fix: remove edgy.nvim to keep window size when open quickfix.

ayamir:
This PR removes `edgy.nvim` b/c it will disrupt the splited window size when term or sidebar openned and the nvim-bpf won't work with it. Besides, readd `gto` with `Telescope lsp_document_symbols` to find symbols.
The previous patch to prevent horizontal toggleterm from obscuring the nvim-tree will fail and cause error so I doesn't add it.

Effect comparison:

Before:

https://github.com/user-attachments/assets/6f701e4e-0705-4c5d-9fa7-201a54603d33

Now:

https://github.com/user-attachments/assets/7ce0883f-efee-4f6e-a095-481f92f85f05


ayamir:
@CharlesChiuGit Please review it when you have time~

CharlesChiuGit:
sure, i will! i'll review it on this weekend~

CharlesChiuGit (approved):
LGTM!

lua/keymap/tool.lua
```diff
@@ -9,12 +9,7 @@ require("keymap.helpers")
 local mappings = {
 	plugins = {
 		-- Plugin: edgy
```
CharlesChiuGit, line 11: i think we can move this keymap to `Plugin: nvim-tree`

CharlesChiuGit (changes requested):

CharlesChiuGit (approved):
LGTM

Issue #1461, opened 2025-05-07
LSP: mason-lspconfig breaking change in v2


fioncat:
The `setup_handlers` was removed in the latest mason-lspconfig (v2.0.0).

See: https://github.com/mason-org/mason-lspconfig.nvim/commit/80f2cd7734ff21da4cf1489c9695e440cdcd139f#diff-7c272e96c20d89744c6832ef1e6bca2f44ee64dc7598dd55bc025ca020b2c9dc

The way to configure lsp may need to change.

ayamir:
https://github.com/mason-org/mason-lspconfig.nvim/issues/545

PR #1466, opened 2025-05-13
refactor(mason): support for mason and mason-lspconfig v2

misumisumi:
- fix ayamir/nvimdots#1461

Having reviewed the [release note](https://github.com/mason-org/mason-lspconfig.nvim/releases/tag/v2.0.0), it appears we can support the new API without significant changes to the existing code. 
This provides a base for considering v2 support.

Currently, we provide our own handlers, but we could leverage [Neovim's](https://neovim.io/doc/user/lsp.html) [built-in mechanism](https://neovim.io/doc/user/options.html#after-directory) for this.
Specifically, we can create a file at `after/lsp/<lsp-name>.lua` within the project root. In my local environment, this allowed me to override the LSP settings.
However, this approach might make it more difficult to separate user configurations from the repository's defaults.

### DIFF
--- a/lua/modules/configs/completion/mason-lspconfig.lua
+++ b/lua/modules/configs/completion/mason-lspconfig.lua
@@ -3,13 +3,11 @@ local M = {}
 M.setup = function()
 	local diagnostics_virtual_lines = require("core.settings").diagnostics_virtual_lines
 	local diagnostics_level = require("core.settings").diagnostics_level
+	local lsp_deps = require("core.settings").lsp_deps
 
-	local nvim_lsp = require("lspconfig")
-	local mason_lspconfig = require("mason-lspconfig")
 	require("lspconfig.ui.windows").default_options.border = "rounded"
-
 	require("modules.utils").load_plugin("mason-lspconfig", {
-		ensure_installed = require("core.settings").lsp_deps,
+		ensure_installed = lsp_deps,
 	})
 
 	vim.diagnostic.config({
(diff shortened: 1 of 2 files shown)


ayamir:
Seems this config won't work? luals can't be autostarted now.

misumisumi:
It can be started in my environment.
Did you update `mazon.nvim` and `mason-lspconfig.nvim`?

misumisumi:
Ah, I understand. There was a cache left in my configuration. It didn't work on a new installation...

misumisumi:
Does this work?

ayamir:
Yes, it works now. Does current impl not support customized lsp config in previous way?

misumisumi:
I just changed it to use the native API instead of configuring it from mason-lspconfig.
We can override the original configuration in vim.lsp.config.
The format doesn't seem to have changed, so I think it's fine to change this part alone.

refs:
- https://neovim.io/doc/user/lsp.html#vim.lsp.config()
- https://github.com/neovim/nvim-lspconfig/blob/master/doc/configs.md

ayamir (approved):

lua/modules/configs/completion/mason-lspconfig.lua
```diff
@@ -88,9 +89,12 @@ please REMOVE your LSP configuration (rust_analyzer.lua) from the `servers` dire
 				{ title = "nvim-lspconfig" }
 			)
 		end
+		return
```
CharlesChiuGit, line 92: this `return` seems reductant?

PR #1489, opened 2025-07-28
feat(python): use pyrefly and ruff to replace pylsp

Cloud0310:
Using pyrelfy to replace the default pylsp as the lsp, using ruff as the linter and formatter.


CharlesChiuGit:
i'm wondering why pyrelfy instead of [ty](https://github.com/astral-sh/ty).
fyi. ty and ruff are from the same company, i think they will work better together.

CharlesChiuGit:
tbh, i think we can close this PR now and wait 'til they got more mature

Cloud0310:
Also, I'm thinking about whether the LSP config should be aligned with original `pylsp` config.
In the last commit, the config is just the default config with all three plugins `pyrefly`, `ty` and `ruff`.

misumisumi:
Sorry I'm so busy that I haven't been able to participate much.

Personally, I don't know if I need to change it from pylsp (I've never used pyrefly but...).
Do we need to switch from pylsp+ruff quickly?

Pylsp is sufficient for my develop environment
Furthermore, I am concerned that if exceptional handling is permitted in a particular language, it will lead to the possibility that the introduction of unstable LSPs will also be permitted in other languages in the future.
I think this could lead to bloating of the cord.

For now, I would like to make a proposal to put it on hold for discussions, projects, etc.

Cloud0310:
I would like to open the pr again when its more suitable to swtich to new python lsp, closing this now.

PR #1475, opened 2025-06-01
refactor: clean up and improve plugin configuration and docs

Jint-lzxy:
Hiiiii srrry for ghosting for a bit things have been a little hectic on my end lately. Since we moved to `0.11` recently, I figured it'd be a good time to clean up the codebase a bit ~~(also gives me a chance to slowly catch up while going thru the files lol)~~ still working my way thru things, so pls bear w me a little longer.
***
Also going slightly off the tangent here, but did we actually remove `main` as the default branch? I tried to pull and git told me `HEAD` is dangling, not sure if that was intentional...? If so, what's the rationale (apologies if I missed the context, I did try searching but didn't come across anything helpful 😭) just wondering if it might cause issues for ppl who have configs pointing to `main`. Maybe it makes more sense to keep `main` on `0.11` and open a `0.12` branch for nightly like we've usually done?

PR #1494, opened 2025-08-01
docs!: update badges and branch info for main and nightly

Jint-lzxy:
Sorry for opening this a lil early! I think this should ideally be merged after the other PRs because it feels more like a breaking change to me. The main reasoning behind this update is outlined [here](https://github.com/ayamir/nvimdots/pull/1475#issuecomment-2926910296), but one important thing is that it will require @ayamir to update the default branch back to `main`.

Also I tweaked the badge style again hopefully it looks a bit nicer now:

new-badges

Would love to hear what yall think! Always happy to hear ideas or suggestions for improvements lol


CharlesChiuGit (approved):
the badge looks nice!

ayamir (approved):

PR #1495, opened 2025-08-01
refactor(lualine): inline chat_progress and improve LSP progress display

Jint-lzxy:
iirc (correct me if im wrong) the current setup would potentially break user configs mainly cuz users can only override a config by creating a file with the same name. If we keep calling the file where `load_plugin` lives `init.lua`, then users would have to create another `init.lua`, which is kinda counterintuitive.

Also since LSP progress updates pretty frequently, i tried to optimize it a bit (e.g., avoiding unnecessary C calls wherever possible). Given that change, i figured it made sense to inline the chat_progress module too. Feels more consistent with how all our other components are set up anyway lol.

### DIFF
--- a/lua/modules/configs/ui/lualine.lua
+++ b/lua/modules/configs/ui/lualine.lua
@@ -250,7 +250,7 @@ return function()
 
 				return function()
 					if not processing then
-						return nil
+						return ""
 					end
 					animation_idx = animation_idx % #animate_chars + 1
 					return string.format("%s %s", icons.aichat.Copilot, animate_chars[animation_idx])


lua/modules/configs/ui/lualine.lua
```diff
@@ -217,13 +228,38 @@ return function()
...
 			cond = conditionals.has_enough_room,
 		},
 
+		chat_progress = {
+			(function()
+				local processing = false
+				local animate_chars = { "⠋", "⠙", "⠹", "⠸", "⠼", "⠴", "⠦", "⠧", "⠇", "⠏" }
+				local animation_idx = 1
+				vim.api.nvim_create_autocmd("User", {
+					pattern = { "CodeCompanionRequestStarted", "CodeCompanionRequestFinished" },
```
Jint-lzxy, line 244: Also regarding this part im not _entirely_ sure why the original impl used a wildcard pattern (`CodeCompanionRequest*`) instead of listing the files explicitly? From what ive read listing them directly is generally faster? Apologies if ive misunderstood anything; im still getting familiar w how this whole code companion system works too lol

ayamir: Use `*` maybe just for simplicity.

PR #1499, opened 2025-08-03
refactor(cmp): streamline copilot integration

Jint-lzxy:


lua/modules/plugins/completion.lua
```diff
@@ -64,19 +63,18 @@ completion["hrsh7th/nvim-cmp"] = {
...
-		lazy = true,
-		cmd = "Copilot",
-		event = "InsertEnter",
-		config = require("completion.copilot"),
-		dependencies = {
-			{
-				"zbirenbaum/copilot-cmp",
-				config = require("completion.copilot-cmp"),
-			},
+completion["zbirenbaum/copilot.lua"] = {
```
ayamir, line 66: This will cause unnecessary download for who doesn't enable copilot.

ayamir (changes requested):

lua/modules/plugins/completion.lua
Jint-lzxy: hmmm i dont think it will (just tested it)? like it wont remove the plugin if the user already has it installed, which imo makes more sense here cuz reauthenticating after like `:Lazy clean` or whatever would be annoying (plus iiuc that [lines up w the docs](https://lazy.folke.io/spec#spec-loading) too)

PR #1502, opened 2025-08-03
feat(mason)!: migrate to v2.x

Jint-lzxy:
Just wrapped up going thru all the changes in mason 2.x (and nvim 0.11 too)! This should be a follow up to #1466 and (hopefully) fix all the issues related to the updated LSP API (like servers not spawning and other weirdness) and works together w #1496.

Acc the main issue was that some mason package names don't match their names in lspconfig, so just using `lsp_deps` directly works for most, but breaks a bunch of others (tldr: that mismatch is what was causing all the edge cases and broken behavior).

I also went ahead and enabled `automatic_installation` because why not lol means we don't have to manually call `vim.lsp.enable()` for everything, which might save a tiny bit of perf too.

Didn't test this super thoroughly yet so pls lmk if anything's still broken!


ayamir:
What's the relation of this pr with #1496, does it require #1496 be merged first?

Jint-lzxy:
> What's the relation of this pr with https://github.com/ayamir/nvimdots/pull/1496, does it require https://github.com/ayamir/nvimdots/pull/1496 be merged first?

these two kinda go hand in hand i think...? not rly dependent on each other but like one fixes `lspconfig` and the other fixes `mason`, and we need both to get completion working again (technically tho merging #1496 first makes more sense cuz it reverts the old workarounds that werent the real fix according to the docs)

CharlesChiuGit (approved):
slightly tested, seems working in most cases.

PR #1516, opened 2025-08-30
fix: use realpath of vim related path

ayamir:
This pr use realpath to replace vim related path to avoid `--no-ignore` failure b/c we use `vim_path` to compare with `vim.uv.cwd`.


CharlesChiuGit (approved):
LGTM

PR #1522, opened 2025-09-17
fix: only use realpath for vim_path

### LINKED ISSUE #1521: [Bug] Startup Error: attempt to concatenate field 'cache_dir' (a nil value)
### Version confirmation

- [x] Confirm

### Following prerequisites

- [x] Confirm

### Not a user config issue

- [x] Confirm

### Neovim version

NVIM v0.11.1 Build type: Release LuaJIT 2.1.1744318430 Run "nvim -V1 -v" for more info

### Operating system/version

macOS 15.6.1

### Terminal name/version

fish

_No response_

### Branch info

main (Default/Latest)

### Fetch Preferences

SSH (use_ssh = true)

Execute the official nvimdots installation script:
bash
if command -v curl >/dev/null 2>&1; then
    bash -c "$(curl -fsSL https://raw.githubusercontent.com/ayamir/nvimdots/HEAD/scripts/install.sh)"
else
    bash -c "$(wget -O- https://raw.githubusercontent.com/ayamir/nvimdots/HEAD/scripts/install.sh)"
fi

Follow the installation prompts (use HTTPS for plugin fetching, shallow clone, etc.).
After installation completes, run nvim in the terminal.
The error above is displayed immediately on startup.

### Expected behavior

Neovim should start normally without cache_dir-related errors, as cache_dir should be properly initialized by the nvimdots configuration.

### Actual behavior

Startup fails due to cache_dir being nil (uninitialized), which causes a Lua concatenation error in lua/core/init.lua:7 (within the createdir function).

### Additional information

To prevent this issue for new users, consider one of the following fixes in the nvimdots codebase:

Auto-create the cache directory in the installation script (e.g., add mkdir -p $(vim -e -c 'echo stdpath("cache")' -c 'quit')/nvim to scripts/install.sh), ensuring the path exists before Neovim starts.
Modify global.lua to handle non-existent paths (e.g., fall back to vim.fn.stdpath("cache") if realpath() returns nil):
lua
self.cache_dir = realpath(vim.fn.stdpath("cache")) or vim.fn.stdpath("cache") .. "/nvim"

Add a note in the installation docs (e.g., Prerequisites wiki) to remind users to create the cache directory if they encounter this error.

Let me know if additional logs or details are needed! Thanks for maintaining this great Neovim config 😊

ayamir:
This pr will fix #1521. In fact, only `vim_path` is needed to be a realpath.

### DIFF
--- a/lua/core/global.lua
+++ b/lua/core/global.lua
@@ -8,8 +8,8 @@ function global:load_variables()
 	self.is_windows = os_name == "Windows_NT"
 	self.is_wsl = vim.fn.has("wsl") == 1
 	self.vim_path = realpath(vim.fn.stdpath("config"))
-	self.cache_dir = realpath(vim.fn.stdpath("cache"))
-	self.data_dir = string.format("%s/site/", realpath(vim.fn.stdpath("data")))
+	self.cache_dir = vim.fn.stdpath("cache")
+	self.data_dir = string.format("%s/site/", vim.fn.stdpath("data"))
 	self.modules_dir = self.vim_path .. "/modules"
 	self.home = self.is_windows and vim.env.USERPROFILE or vim.env.HOME
 end


ayamir:
> LGTM,
> 
> tbh i dont really know what `vim.uv.fs_realpath` does since the [neovim doc](https://neovim.io/doc/user/luvref.html#uv.fs_realpath()) doesn't say much about it lol

It converts symlink and relative path into absolute path.

PR #1584, opened 2026-06-30
fix(ui): guard against nil colors_name in lualine/bufferline

charliie-dev:
## What

`lualine.lua` and `bufferline.lua` call `vim.g.colors_name:find("catppuccin")` directly. `vim.g.colors_name` is `nil` until a colorscheme has been applied (e.g. `nvim --clean`, or before `:colorscheme` runs), so `nil:find()` raises `attempt to index a nil value` if either module loads at that point.

This coalesces to an empty string in all three spots:

- `lua/modules/configs/ui/lualine.lua:44`
- `lua/modules/configs/ui/lualine.lua:60` (inside the `ColorScheme` autocmd callback)
- `lua/modules/configs/ui/bufferline.lua:45`

```lua
(vim.g.colors_name or ""):find("catppuccin")
```

## Why

This matches the nil-safe pattern already used in `lua/modules/utils/init.lua` (`get_palette` and `set_global_hl`), so it just brings these two UI configs in line with the rest of the codebase.

Behaviour is unchanged when a colorscheme is set:
- `("catppuccin-mocha" or ""):find("catppuccin")` → match (truthy), as before
- `(nil or ""):find("catppuccin")` → no match, **no error** (previously errored)

## Verification

- `stylua lua/` — clean.
- `grep -rn 'vim.g.colors_name:find' lua/` → no bare occurrences remain.
- Verified empirically that `vim.g.colors_name` is `nil` under `nvim --clean`.

### DIFF
--- a/lua/modules/configs/ui/bufferline.lua
+++ b/lua/modules/configs/ui/bufferline.lua
@@ -42,7 +42,7 @@ return function()
 		highlights = {},
 	}
 
-	if vim.g.colors_name:find("catppuccin") then
+	if (vim.g.colors_name or ""):find("catppuccin") then
 		local cp = require("modules.utils").get_palette() -- Get the palette.
 
 		local catppuccin_hl_overwrite = {
--- a/lua/modules/configs/ui/lualine.lua
+++ b/lua/modules/configs/ui/lualine.lua
@@ -41,7 +41,7 @@ vim.api.nvim_create_autocmd("LspProgress", {
 })
 
 return function()
-	local has_catppuccin = vim.g.colors_name:find("catppuccin") ~= nil
+	local has_catppuccin = (vim.g.colors_name or ""):find("catppuccin") ~= nil
 	local colors = require("modules.utils").get_palette()
 	local icons = {
 		diagnostics = require("modules.utils.icons").get("diagnostics", true),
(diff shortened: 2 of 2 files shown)


hiSandog:
The nil guard looks right for early startup. One small edge case worth covering is the `ColorScheme` autocmd path too: if the callback fires while `vim.g.colors_name` is unset or temporarily cleared, both lualine and bufferline should keep using the neutral branch rather than throwing. A tiny startup smoke test with no colorscheme applied would document the intended behavior.

charliie-dev:
Thanks @hiSandog ! Good news — the `ColorScheme` autocmd path is already covered by this PR, and I traced every `colors_name` access reachable from a `ColorScheme` autocmd to be sure.

There are two `ColorScheme` autocmds in the config, and both are safe:

1. **lualine's** (`LualineColorScheme`, in `custom_theme()`): its callback line is exactly one of the three guards in this PR —
   ```lua
   -  has_catppuccin = vim.g.colors_name:find("catppuccin") ~= nil
   +  has_catppuccin = (vim.g.colors_name or ""):find("catppuccin") ~= nil
   ```
   so if it fires with `colors_name` unset it takes the neutral (`"auto"` / `gen_hl -> nil`) branch instead of throwing.

2. **the palette-reinit one** (`__builtin_palette`, in `modules/utils/init.lua`): its callback goes through `init_palette()` and `set_global_hl()`, both of which already use `(vim.g.colors_name or "")` — so that path was never affected.

bufferline has no `ColorScheme` autocmd (it's configured once), and its single check is guarded here too.

Quick empirical check of the guarded expression with no colorscheme applied:
```
$ nvim --clean --headless -c 'lua local n=nil; print(pcall(function() return (n or ""):find("catppuccin") ~= nil end))' -c 'qa!'
true    false      -- pcall ok = true (no throw), has_catppuccin = false (neutral branch)
```

On the smoke test: this repo has no Lua test harness (CI is stylua + luacheck only), so rather than pull one in for a single case I've documented the intended behavior with the check above. Happy to add more detail if useful.

ayamir (approved):

PR #1591, opened 2026-07-26
feat(ui): add nvchad ui mode

ayamir:
This PR introduces nvchad colorscheme and ui components for nvimdots.

## Summary
- add a nvchad colorscheme entry plus nvconfig compatibility layer for Base46 themes without using chadrc/volt
- use nvchad-ui components in nvchad mode while keeping bufferline/lualine for non-NvChad themes
- preserve nvimdots buffer/tab keymaps across both UI modes and remove alpha, toggleterm, and nvim-bufdel usage


ayamir:
The direction is introducing nvchad as a way to be parallel with existing UI settings, but IMO it's better to use nvchad to replace existing UI settings, maintaining two modes is somewhat cumbersome.

ayamir:
Another changes:
1. replace lots plugins with snacks.nvim
2. use improved search.nvim for snacks.nvim


https://github.com/user-attachments/assets/981fe438-3242-41df-a185-e6c52fea6353

